Issue #13, opened 2022-02-11
cloud-init: homedir and shell are required when creating new users


mudler:
Booting with the following `cloud-init` file fails adding the vagrant user:

```yaml
#cloud-config

# Add additional users or set the password/ssh keys for root
users:
- name: "root"
  passwd: "ros"
  ssh_authorized_keys:
  - https://raw.githubusercontent.com/hashicorp/vagrant/main/keys/vagrant.pub
- name: "vagrant"
  passwd: "vagrant"
  ssh_authorized_keys:
  - https://raw.githubusercontent.com/hashicorp/vagrant/main/keys/vagrant.pub
```

`cos-setup-network`:
![Screenshot from 2022-02-11 10-04-16](https://user-images.githubusercontent.com/2420543/153564290-af28387a-c54b-41e9-b5ea-11697ecb56cf.png)

mudler:
It actually needs a bump in elemental-cli :smiling_face_with_tear:

PR #14, opened 2022-02-11
Cloud init fixups

mudler:
This PR is multi-fold and contains several fixups I've found while digging cloud-init issues. The changes are split into separate commit for a better review. 

It introduces some logical changes, while until now all other changes were pretty cosmetical:

- ros-installer by default reading /oem/userdata for automated installation
- services have been reworked to wait on cos-setup-network rather network-online. This ensures userdata is loaded so installation configuration is picked up correctly
- cloud-init datasources logic changed slightly: in order to be sure that we pull the data from the network, we call explictly the userdata fetched if none was found previously. This impact only first-boot as subsequent boots have the userdata pulled correctly. I would suggest in the long term to encode cloud-init files directly in the `network` stage to be sure we have the network. But since this is optional and at our discretion, this PR enforces this behavior, see https://github.com/rancher-sandbox/cOS-toolkit/issues/1124 for a follow-up.

Fixes: https://github.com/rancher-sandbox/os2/issues/9
Fixes: https://github.com/rancher-sandbox/os2/issues/8
Fixes: https://github.com/rancher/os2/issues/14

### DIFF
--- a/cmd/ros-installer/main.go
+++ b/cmd/ros-installer/main.go
@@ -14,7 +14,7 @@ import (
 var (
 	automatic   = flag.Bool("automatic", false, "Check for and run automatic installation")
 	printConfig = flag.Bool("print-config", false, "Print effective configuration and exit")
-	configFile  = flag.String("config-file", "", "Config file to use, local file or http/tftp URL")
+	configFile  = flag.String("config-file", "/oem/userdata", "Config file to use, local file or http/tftp URL")
 	powerOff    = flag.Bool("power-off", false, "Power off after installation")
 	yes         = flag.Bool("y", false, "Do not prompt for questions")
 )
--- a/framework/files/etc/systemd/system/ros-installer.service.d/override.conf
+++ b/framework/files/etc/systemd/system/ros-installer.service.d/override.conf
@@ -1,0 +1,3 @@
+[Unit]
+After=
+After=cos-setup-network.service
--- a/framework/files/system/oem/01_ros-rootfs.yaml
+++ b/framework/files/system/oem/01_ros-rootfs.yaml
@@ -52,7 +52,6 @@ stages:
         device:
           label: COS_PERSISTENT
         expand_partition:
-          size: 0
   network:
     - <<: *datasource
     # Trigger /oem/userdata if was fetched late
--- a/framework/files/system/oem/01_ros-rootfs.yaml
+++ b/framework/files/system/oem/01_ros-rootfs.yaml
@@ -42,6 +42,9 @@ stages:
     # Ensures that if no userdata is found at this stage,
     # we retry later and run the boot hook
     # This is relevant only on first-boot
+    # Mind that userdata can be in the standard cloud-init syntax.
+    # If we don't find any userdata at this stage (because no network is present)
+    # We source the respective fields afterwards
     - if: '[ ! -f /oem/userdata ]'
       files:
       - path: /oem/userdata_load
--- a/framework/files/system/oem/01_ros-rootfs.yaml
+++ b/framework/files/system/oem/01_ros-rootfs.yaml
@@ -34,6 +34,9 @@ stages:
           /var/lib/cni
         PERSISTENT_STATE_BIND: "true"
   rootfs.before:
+    # Try to get network before trying to fetch datasource from the net
+    - commands:
+      - wicked ifup eth0
     - &datasource
       name: "Pull data from provider"
       datasource:
--- a/framework/files/system/oem/01_ros-rootfs.yaml
+++ b/framework/files/system/oem/01_ros-rootfs.yaml
@@ -34,10 +34,17 @@ stages:
           /var/lib/cni
         PERSISTENT_STATE_BIND: "true"
   rootfs.before:
-    - name: "Pull data from provider"
+    - &datasource
+      name: "Pull data from provider"
       datasource:
-        providers: ["cdrom"]
+        providers: ["cdrom", "gcp", "openstack", "aws", "azure", "hetzner", "packet", "scaleway", "vultr", "digitalocean", "metaldata" ]
         path: "/oem"
+    # Ensures that if no userdata is found at this stage,
+    # we retry later and run the boot hook
+    # This is relevant only on first-boot
+    - if: '[ ! -f /oem/userdata ]'
+      files:
+      - path: /oem/userdata_load
   rootfs.after:
     - if: '[ ! -f /run/cos/recovery_mode ] && [ ! -f /run/cos/live_mode ]'
       name: "Grow persistent"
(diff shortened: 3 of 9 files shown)


tests/assets/cloud_init.yaml
```diff
@@ -1,30 +1,14 @@
...
+
+# Add additional users or set the password/ssh keys for root
+users:
+- name: "root"
+  passwd: "ros"
+  ssh_authorized_keys:
+  - https://raw.githubusercontent.com/hashicorp/vagrant/main/keys/vagrant.pub
+- name: "vagrant"
+  passwd: "vagrant"
+  shell: "/bin/bash"
```
mudler, line 11: Note, shell and homedir _are_ required or the user results invalid for the system. See #13

mudler: I've added it to the docs here: https://github.com/rancher-sandbox/os2/pull/14/commits/1cdc879254dbf9e481ee257b48568b9c026adbc6 for the time being

framework/files/etc/systemd/system/ros-installer.service.d/override.conf
```diff
@@ -0,0 +1,3 @@
+[Unit]
+After=
```
Itxaka, line 2: Why is there an empty after?

Itxaka (approved):

framework/files/etc/systemd/system/ros-installer.service.d/override.conf
mudler: This is to clear out whatever is defined from the target service and avoid any potential issue. 

I've been struggling with this usually with `ExecPre`s when overriding units, some of them don't accept multiple values, and yeah, things kinda gets tricky.

cmd/ros-installer/main.go
```diff
@@ -14,7 +14,7 @@ import (
 var (
 	automatic   = flag.Bool("automatic", false, "Check for and run automatic installation")
 	printConfig = flag.Bool("print-config", false, "Print effective configuration and exit")
-	configFile  = flag.String("config-file", "", "Config file to use, local file or http/tftp URL")
+	configFile  = flag.String("config-file", "/oem/userdata", "Config file to use, local file or http/tftp URL")
```
davidcassany, line 17: I am curious about that. What's the content of `/oem/userdata`? If this is created by the datasource yip plugin and then loaded again as a yip config file I think it should be  `/oem/userdata.yaml`.

framework/files/system/oem/01_ros-rootfs.yaml
```diff
@@ -34,23 +34,34 @@ stages:
...
-      datasource:
-        providers: ["aws", "gcp", "openstack", "cdrom"]
-        path: "/oem"
+    - <<: *datasource
+    # Trigger /oem/userdata if was fetched late
+    # when network was available
+    # This is relevant only on first-boot
+    - if: '[ -f /oem/userdata ] && [ -f /oem/userdata_load ]'
+      commands:
+      - elemental cloud-init -s initramfs /oem/userdata
```
davidcassany, line 62: This is odd... why calling initramfs stage during network stage?

I think it is wrong if userdata includes an initramfs stage that can be called at network stage, why not just make the userdata include a `network.after` stage?

davidcassany (commented):
I find the userdata stages management confusing... couldn't just `initramfs` and `rootfs` stages of the userdata simply be moved to `network.after`?

mudler:
> I find the userdata stages management confusing... couldn't just `initramfs` and `rootfs` stages of the userdata simply be moved to `network.after`?

The `userdata` might be in cloud-init format, so there you have automatically mapped them at `boot` and `initramfs` stage. So we can't directly hook on that

cmd/ros-installer/main.go
mudler: this is expected to be the user input only without any further processing. I've checked the code and `userdata.yaml` and `userdata` are equivalent (there is no conversion happening, while I was expecting it on the first place too)

framework/files/system/oem/01_ros-rootfs.yaml
mudler: so the problem relies on the fact that the input is in standard cloud-init format, and not in our extended syntax.

When `yip` reads cloud-inits files in the standard syntax it encodes them in the `boot` stage for user/files/etc and `initramfs` for setting up the hostname. I do believe too that we should move this encoding simply to `network`, but on the other end that implies that the cloud-init content wouldn't run if there is no network (thus, it wouldn't allow us to setup networking)

mudler: I'd rather open a separate issue to make `yip` consistent and encode everything under `boot`, but moving it to `network` would leave any network settings via cloud-init in the cold

mudler: I've opened https://github.com/rancher-sandbox/cOS-toolkit/issues/1124 for further discussion, because it has different implications. IF for example, we don't encode the hostname into initramfs, any other stage that might depend on the hostname might react differently, as `boot` might trigger _too late_

davidcassany: Oh I see this is even more convoluted that what I initially thought 😅 

IMHO the solution would be not using cloud-init format... I am not sure how feasible this is. Are we using our custom cloud-init formatted config or are we pulling it form somewhere else? I also understand that using cloud-init format can be beneficial to ease creating a community around it, using already known tools/syntaxes helps.

cmd/ros-installer/main.go
davidcassany: There is no conversion. The only difference is that  `userdata.yaml` is granted to be readable by yip (extended or not). But probably this doesn't make a difference here.

davidcassany (approved):

framework/files/system/oem/01_ros-rootfs.yaml
mudler: That's the issue indeed.. it is pulled from outside as user input, so this is totally user-facing in a variety of ways..

For example, rancherd configuration is _embedded_ into the cloud init one, so the user input is one single file with all the config needed, that is system+rancherd. 

I'm indeed in favor of our syntax which is much more flexible.. but "forcing" our syntax might be too much from the user perspective, and os2 is mostly documented only around the standard syntax. 

On the other way around, scoping it entirely on the standard syntax might also alleviate us to not have to care to support the extended syntax via datasources, but rather just that available for who is customizing os2 from a system level.

mudler: This solution has some caveats in any case too. Because _if_ datasources pulls very late and we specify an hostname from there, the hostname is set anyway lower in the stack..

mudler:
been hitting https://github.com/rancher-sandbox/os2/issues/15, adding a simple workaround, but keeping still the cloud-init standard syntax to have a more realistic scenario on our hands

Issue #20, opened 2022-02-17
TPM emulation


mudler:
Follow-up of #7 

- [x] https://github.com/rancher-sandbox/os2/pull/21
- [ ] Allow to enable via configuration TPM emulation which is supported in https://github.com/rancher-sandbox/go-tpm. See https://github.com/rancher-sandbox/go-tpm/blob/main/get_test.go#L105 as an example.
- [ ] Draft a new os2 release - the feature above helps testing

PR #21, opened 2022-02-17
Consume go-tpm in place of tpm in rancherd/os2

mudler:
Part of #20. Will allow later on to set an emulated device.

Those are the changeset on top of `tpm` that was provided on rancherd: https://github.com/rancher-sandbox/go-tpm/compare/c2203f40dd8de3edc9dfe4765d01991c79231c73...main , in summary:

- Imported some functions that were used in os2, but relevant to tpm only, so that's why the huge deletion in this changeset
- Common code has been refactored so could be used both in os2 and in rancherd
- Added tests spawning a ws server mimicking what kubernetes does when challenging nodes
- Added TPM emulation capabilities (not yet exposed in os2 configs)


go.mod
```diff
@@ -33,17 +33,17 @@ replace (
...
-	github.com/onsi/ginkgo/v2 v2.1.1
+	github.com/onsi/ginkgo/v2 v2.1.3
 	github.com/onsi/gomega v1.18.1
 	github.com/pin/tftp v2.1.0+incompatible // indirect
 	github.com/pkg/errors v0.9.1
+	github.com/rancher-sandbox/go-tpm v0.0.0-20220217133323-a7b15ad1f8f7
 	github.com/rancher/fleet/pkg/apis v0.0.0-20210927195558-4aaa778d23dd
 	github.com/rancher/lasso v0.0.0-20210709145333-6c6cd7fd6607
 	github.com/rancher/rancher/pkg/apis v0.0.0-20211013185633-a636bda2a00e
-	github.com/rancher/rancherd v0.0.1-alpha9.0.20211028172625-bdf5642d62d5
```
mudler, line 46: and this unties us from rancherd. I would actually vote to have rancherd using go-tpm as it encompasses the general usage done in both cases.

Itxaka (approved):
looks good, would need to check the go-tpm changes to understand more about this :D

mudler:
> looks good, would need to check the go-tpm changes to understand more about this :D

yep without a bit of exploring how tpm work is hard, but this is mostly refactoring and adding options around so can be re-used and tested. The emulation part is the real "new" addition, which is used in the unit tests

Itxaka:
yup, I already had a look yesterday to the branch...could not understand much :P but looked kind of straighforward

PR #22, opened 2022-02-22
Update cos

mudler:


mudler:
Changes should be more or less in this direction. I'm just trying to get now CI green as there were couple of things to deal with cloud init. For example we were consuming cOS defaults which were clashing with os2, so hence why few small additions. Actually we override most of the config, so we don't need default examples at all.

framework/files/system/oem/00_recovery.yaml
```diff
@@ -0,0 +1,23 @@
+name: "Recovery partition boot setup"
+stages:
+   rootfs.before:
+     - if: |
```
davidcassany, line 4: I am wondering if this logic should be part of the initrd logic. I'd say it makes sense identifying at very early stages the mode we are in and set the sentinel flag before any stage. Having this as a core feature is really helpful for cloud-init purposes and prevents the user of having to deal with this non obvious setups. What do you think?

I believe it would be simple to implement that as part of the default initrd logic.

framework/files/system/oem/01_ros-rootfs.yaml
```diff
@@ -8,6 +8,17 @@ stages:
           echo rancher-${RANDOM} > /usr/local/etc/hostname
         fi
         ln -sf /usr/local/etc/hostname /etc/hostname
+    - if: '[ ! -f "/run/cos/recovery_mode" ]'
+      name: "Persist /etc/machine-id"
```
davidcassany, line 12: Like above we could think up of setting persistent files in some sort of generic approach, even I do not see it as obvious as the boot mode sentinel file.

Persisting files/folder can easily require some additional custom configs or tweaks.

davidcassany (approved):
Looks good :+1:

Just added comments with ideas to slightly improve cOS and smooth derivatives setup

framework/files/system/oem/00_recovery.yaml
mudler: I was thinking on the same lines too, but I always wondered then how could be like having support for custom partition labels, would end up the initrd to be strictly tied to the underlying configuration of the OS ( system labels set in the config file ), while in this way it's completely dynamic.

Maybe a middle ground would be to provide syntax sugar directly in the cloud-init with some form, or even split all the defaults cloud-init in `cOS` in single packages. I didn't wanted to dup here, and would have been nice If I could just `luet install -y cloud-config/sentinels` additionally from the `cos-core` package (which doesn't ship any cloud-configs by default)

mudler:
got a green build here https://github.com/rancher-sandbox/os2/actions/runs/1887262949, rebased to add a comment on the cloud-init file with the created Issues

framework/files/system/oem/00_recovery.yaml
davidcassany: > I was thinking on the same lines too, but I always wondered then how could be like having support for custom partition labels, would end up the initrd to be strictly tied to the underlying configuration of the OS ( system labels set in the config file ), while in this way it's completely dynamic. I'm just unsure on how complex it could became hitting that road

Note this is already the case, unfortunately. Labels are only configurable and customizable at build time currently. Probably this is something to revisit. IMHO, labels should be dynamic at install time, we are not there yet, there is grub static configuration files and few other artifacts that are depending on specific label names. Yes, it sucks. I did some little brainstorming and investigations here https://github.com/rancher-sandbox/cOS-toolkit/issues/160#issuecomment-977988331

@mudler do you think it makes sense to make a new issue out of it?

mudler: > > I was thinking on the same lines too, but I always wondered then how could be like having support for custom partition labels, would end up the initrd to be strictly tied to the underlying configuration of the OS ( system labels set in the config file ), while in this way it's completely dynamic. I'm just unsure on how complex it could became hitting that road
> 

Yes correct, few parts are dynamic and few parts aren't (like installer sourcing configs from runtime), that's why I wouldn't push much more into static if we are going to pursue a dynamic approach in the long-run in detecting partitions, etc.

> Note this is already the case, unfortunately. Labels are only configurable and customizable at build time currently. Probably this is something to revisit. IMHO, labels should be dynamic at install time, we are not there yet, there are grub static configuration files and few other artifacts that are depending on specific label names. Yes, it sucks. I did some little brainstorming and investigations here [rancher-sandbox/cOS-toolkit#160 (comment)](https://github.com/rancher-sandbox/cOS-toolkit/issues/160#issuecomment-977988331)

I have a mixed feeling here. Having derivatives sharing same logic pieces alltogether makes debugging environments easily as they look all the same and you expect the same behavior. offering way too many customization pieces might break some of the typical flows (not saying that's bad), but we should define first something all derivatives should adhere to, at least to avoid fragmentation.

> 
> @mudler do you think it makes sense to make a new issue out of it?

Yes, this should be at least clarified, whatever direction we will take on it.

PR #34, opened 2022-03-02
Rework golangci-lint config and github CI

Itxaka:
Update the values on the config file for golangci
Use golangci on github job instead of 4 different binaries, some of them
deprecated (golint)

Enable more linters:
 - dupl for duplicated code
 - goconst for strings that can be constants
 - goheader for checking license headers

Add license headers
Fix some underlying issues like no errors or using outdated libs


pkg/config/read_test.go
```diff
@@ -81,16 +97,18 @@ func WSServer(ctx context.Context, data map[string]interface{}) {
...
 			writer, _ := conn.NextWriter(websocket.BinaryMessage)
-			json.NewEncoder(writer).Encode(data)
+			_ = json.NewEncoder(writer).Encode(data)
 		}
 	})
 
 	s.Handler = m
 
-	go s.ListenAndServe()
+	go func() {
```
Itxaka, line 106: this is a bit strict, especially for tests. Maybe we should skip the ci analysis for test files...

pkg/questions/questions.go
```diff
@@ -137,6 +153,6 @@ func PromptOptional(text, def string) (string, error) {
 			answer = def
 		}
 
-		return answer, nil
+		return answer, nil //nolint:staticcheck
```
Itxaka, line 156: This claims that is returning on the first loop iteration and I agree, but Im not sure I understand why this was done this way and not out of the loop. @mudler any ideas? Would it be safe to remove the loop here?

pkg/util/prompt.go
```diff
@@ -7,7 +23,7 @@ import (
 	"os"
 
 	"github.com/pkg/errors"
-	"golang.org/x/crypto/ssh/terminal"
+	"golang.org/x/term"
```
Itxaka, line 26: "golang.org/x/crypto/ssh/terminal" is deprecated

.github/workflows/static_analisys.yaml
```diff
@@ -1,5 +1,12 @@
 name: Static Analysis
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - master
```
mudler, line 5: why restricting only on those two branches? could be handy while iterating in a specific branch

.github/workflows/static_analisys.yaml
```diff
@@ -12,26 +19,7 @@ jobs:
...
-      run: gocyclo -over 30 .
-
-    - name: Golint
-      run: golint -set_exit_status $(go list -tags ci ./...)
-
-    - name: Staticcheck
-      run: staticcheck -go 1.12 ./...
\ No newline at end of file
+    - name: Analysis
+      uses: golangci/golangci-lint-action@v3
```
mudler, line 23: :+1:

pkg/config/read_test.go
mudler: yuck, well..

pkg/questions/questions.go
mudler: it doesn't return always as if there is no selection it hits `continue`, so it would keep looping. On the other hand I agree, better have a break statement and returning at the end

mudler (approved):
Looks good! thanks for taking care of it!

.github/workflows/static_analisys.yaml
Itxaka: mainly cause its the way on doing it with the fork-and-branch flow that is commonly used...I cant revert it back but that means:
 - We run both tests and analysis twice, one for branch, one for PR
 - We have to work on the original repo, which is risky (`git push --force` while in master.....chaos ensures)

Which is why my preferred method is to use the fork-and-branch. But as we need to use the branch from main repo for tests....it should not matter much, it will just avoid duplicating the jobs. IMHO working on the main repo is risky for me, as Im used to fuck up things in git and re-fork the whoe thing, treating my forked repo as cattle, dont care if it dies, so for me I will continue doing fork-and-branch and then send the branch against the main repo once Im happy with the result so it runs the integration tests.

If you want to iterate over a branch and run those, for the moment they can be run locally and they take about 90 seconds, which is acceptable IMO (and recommended! pushing to just check lint stuff is kind of lazy :P)

Itxaka: We could also just drop the trigger on PR, and just run on branches to avoid the duplication

pkg/questions/questions.go
Itxaka: I think this was just copied from the `Prompt` function above which waits until it gets a proper response, but in this case, any response is valid so it should finish on the first try

.github/workflows/static_analisys.yaml
mudler: > mainly cause its the way on doing it with the fork-and-branch flow that is commonly used...I cant revert it back but that means:
> 
>     * We run both tests and analysis twice, one for branch, one for PR
> 
>     * We have to work on the original repo, which is risky (`git push --force` while in master.....chaos ensures)

Don't really follow here, but I'm fine with this limitation at the end... we could also enforce protection and disable force pushes on master so they get rejected in Github

> 
> 
> Which is why my preferred method is to use the fork-and-branch. But as we need to use the branch from main repo for tests....it should not matter much, it will just avoid duplicating the jobs. IMHO working on the main repo is risky for me, as Im used to fuck up things in git and re-fork the whoe thing, treating my forked repo as cattle, dont care if it dies, so for me I will continue doing fork-and-branch and then send the branch against the main repo once Im happy with the result so it runs the integration tests.

That's ok then, it's enough for me :+1:  better have a flow that suits everyone, I'm ok to open up a PR to have static checks
> 
> If you want to iterate over a branch and run those, for the moment they can be run locally and they take about 90 seconds, which is acceptable IMO (and recommended! pushing to just check lint stuff is kind of lazy :P)

well, it assumes you have the same setup of the CI locally in every machine,  so sometime it's way much easier to check up on GH directly ;)

Itxaka: doesnt really matters anymore, I moved to trigger only on branch push so it should work for master and branches. After all, we want to have integration tests always and we only get them by pushing to the branch so...we need to push to the main repo yes or yes.

PR #36, opened 2022-03-03
Do not nullify the extra data from the config

### LINKED ISSUE #24: extra yaml fields not functional when running manual installation
When an installation is performed manually with ros-installer `--config-file`, `runcmd`, and other extra YAML fields are not preserved:

Input config file:
```yaml
#cloud-config
rancherd:
  role: cluster-init
  rancherValues:
    features:
    - multi-cluster-management=true
ssh_authorized_keys:
  - github:mudler
runcmd:
  - curl -fL https://raw.githubusercontent.com/rancher/rancherd/master/install.sh | sh -
```

After install:

```bash
rancher-8421:/oem # cat 99_custom.yaml 
#cloud-config
rancheros: {}
ssh_authorized_keys:
- github:mudler
```

Seems ros-installer is manipulating the yaml fields in a way that extra fields are not carried over.

Itxaka:
For some reason the installer was assuming that if you dont do an
automatic installation, your "extra data" from the config file should be
gone.

I do not understand why as automatic is not the same as providing a
config file with values which may have nothing to do with the
installation.

This patch removes the nulling of the extra data

Fixes #24 

### DIFF
--- a/pkg/install/install.go
+++ b/pkg/install/install.go
@@ -75,7 +75,6 @@ func runInstall(cfg config.Config, output string) error {
 		if err != nil || !val {
 			return err
 		}
-		cfg.Data = nil
 	}
 
 	if cfg.RancherOS.Install.ConfigURL == "" && !cfg.RancherOS.Install.Automatic {


Itxaka:
ok, somehow this breaks the `ros-installer -automatic` in some weird way, 'cause Im seeing everything installed properly but then it cannot find the `MachineRegistration` kind in `rancheros.cattle.io/v1`

So before this patch, the data provided in the cloud-config would be erased if automatic was set...but if no data is passed...why does it matter? And how that affects the ros-operator, which is the one that installs the  `rancheros.cattle.io/v1` api to not recognize that kind?

any ideas @mudler ? Im setting a test env locally to check manually what is going on but I dont really understand it properly as everything seems to be correctly installed, ros-operator included....

mudler:
It might be also some flakyness in the tests due to timeouts, did you tried retriggering/locally? is that persisting?

Itxaka:
triggered twice, failed twice in the same point. Currently trying locally

mudler:
ah! it's green now :D

mudler (approved):

Itxaka:
Locally also worked....wtf

PR #40, opened 2022-03-04
Use ros-operator image for the ros-operator chart

Itxaka:
Create a simple ros-operator image with only the ros-operator binary on
it as it should be enough.

Use it as default for the chart.

Introduce REPO_ROS_OPERATOR to point to the proper repo to store the
ros-operator.

Set the REPO_ROS_OPERATOR to ros-operator-ci for ci and ros-operator for
release


mudler (approved):
Looks good!

.github/workflows/integration-tests.yaml
```diff
@@ -11,6 +11,7 @@ jobs:
       DOCKER_PASSWORD: ${{ secrets.QUAY_PASSWORD }}
       DOCKER_REGISTRY: quay.io
       REPO: quay.io/costoolkit/os2-ci
+      USE_FRAMEWORK_IMAGE: true # makes it use the $REPO-framework image for the chart
```
mudler, line 14: just a note, we need to set this also in the release pipeline, this is going only to apply for integration tests otherwise

mudler:
not really related, but https://status.quay.io/incidents/rp4pbz2nv3vh might explain today CI failures :face_with_spiral_eyes:

.github/workflows/integration-tests.yaml
Itxaka: currently just setting it for testing, I would like to discuss on monday if we should just create a ros-installer, bin only image instead and make it the default for our chart building here

Itxaka:
Huge difference in image size with this patch, by targeting the framework image on build-framework we got down from 500Mb to 139Mb

mudler:
> Huge difference in image size with this patch, by targeting the framework image on build-framework we got down from 500Mb to 139Mb

I'd say that should be enough for now, we can cut specific image too, but 139 MB sounds reasonable

Makefile
```diff
@@ -29,6 +29,7 @@ build-framework:
 		--build-arg CACHEBUST=${CACHEBUST} \
 		--build-arg IMAGE_TAG=${TAG} \
 		--build-arg IMAGE_REPO=${REPO}-framework \
+		--target framework \
```
mudler, line 32: argh. that was the missing piece -.- good catch!

PR #46, opened 2022-03-10
Add reboot flag

### LINKED ISSUE #45: Server is not automatically reboot after installation
I generated a RancherOS iso with elemental-image-build from [rancher-sandbox/os2-opensuse-image](https://github.com/rancher-sandbox/os2-opensuse-image), then installed a VM with it.
Installation went fine but there was no automatic reboot at the end, and this should be the default behavior.

Moved from  https://github.com/rancher/os2/issues/15

Itxaka:
Fixes #45

### DIFF
--- a/pkg/install/install.go
+++ b/pkg/install/install.go
@@ -28,7 +28,7 @@ import (
 	"sigs.k8s.io/yaml"
 )
 
-func Run(automatic bool, configFile string, powerOff bool, reboot bool, silent bool) error {
+func Run(automatic bool, configFile string, powerOff bool, reboot bool, noRebootAutomatic bool, silent bool) error {
 	cfg, err := config.ReadConfig(context.Background(), configFile, automatic)
 	if err != nil {
 		return err
@@ -46,8 +46,8 @@ func Run(automatic bool, configFile string, powerOff bool, reboot bool, silent b
 		cfg.RancherOS.Install.Automatic = true
 	}
 
-	// If we set the installation to automatic, reboot is set to true
-	if automatic {
+	// If we set the installation to automatic, reboot is set to true unless we set no reboot
+	if automatic && !noRebootAutomatic {
 		cfg.RancherOS.Install.Reboot = true
 	}
 
--- a/pkg/install/install.go
+++ b/pkg/install/install.go
@@ -28,7 +28,7 @@ import (
 	"sigs.k8s.io/yaml"
 )
 
-func Run(automatic bool, configFile string, powerOff bool, silent bool) error {
+func Run(automatic bool, configFile string, powerOff bool, reboot bool, silent bool) error {
 	cfg, err := config.ReadConfig(context.Background(), configFile, automatic)
 	if err != nil {
 		return err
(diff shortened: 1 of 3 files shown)


mudler:
I think that we are missing still something here. Reboot should be default if automatic is set to true, alternatively the user can specify `reboot` to `false` in order to disable this behavior

Itxaka:
> I think that we are missing still something here. Reboot should be default if automatic is set to true, alternatively the user can specify reboot to false in order to disable this behavior

blergh. I will add that. Or redo the whole ros-installer, not sure...

pkg/install/install.go
```diff
@@ -28,24 +28,20 @@ import (
...
-	}
+	cfg.RancherOS.Install.PowerOff = powerOff
+	cfg.RancherOS.Install.Reboot = reboot
+	cfg.RancherOS.Install.Automatic = silent
 
 	if automatic && !cfg.RancherOS.Install.Automatic {
 		return nil
 	}
 
-	if silent {
```
mudler, line 45: I think this construct was there to make sure that the CLI overrides the config only to enable certain aspects, but the config can still be configured independently. That is, if the user runs the installer without silent, or poweroff, the config section takes precedence. In this way it would completely override what's specified in the config

Itxaka: goddammit, everything about this is wroooong

Itxaka:
> Reboot should be default if automatic is set to true, alternatively the user can specify reboot to false in order to disable this behavior

This is kind of against each other...the default is to set reboot to false. But if automatic we set it to true. But the default value is false so how we differentiate from a user passed flag setting it to false?? Another flag?

This is such a mess.

Itxaka:
actually, shouldnt the reboot stuff int he case of the automatic come directly from the config file which contains the data for the automatic installation @mudler ?

Itxaka:
> actually, shouldnt the reboot stuff int he case of the automatic come directly from the config file which contains the data for the automatic installation @mudler ?

Yeah this looks like it, we even check if the config has the automatic value if we set the automatic flag and fail out (well it just exits with no error lol) if not. So the values should be set on the automatic config file provided, not via flag

mudler:
the doc advertizes reboot to be the default behavior for automated installs (that is, for example, provisioned with kubernetes), looks good now indeed!

pkg/install/install.go
```diff
@@ -38,14 +38,23 @@ func Run(automatic bool, configFile string, powerOff bool, silent bool) error {
...
+		cfg.RancherOS.Install.Reboot = true
 	}
 
 	if silent {
 		cfg.RancherOS.Install.Automatic = true
 	}
 
+	// If we set the installation to automatic, reboot is set to true
+	if automatic {
+		cfg.RancherOS.Install.Reboot = true
```
mudler, line 51: mmm no way to set reboot to off if automatic is set?

Itxaka: well..not with the current config load and such. I mean, the flag is gonna be set to false always by default, that is the nature of the bool flags so hwo to discern between the default value and the explicit value set by an user (you can set --reboot=false/true)

I guess we could introduce a `--no-reboot` flag but goddammit, I rather redo the installer in rust  than add more corner cases.

I guess the problem here is a lack of vision/feature set well defined. Like yeah, config file should have more priority, flag should have more priority, with automatic is automatic and you cant change it, etc...


I could also use the question stuff to ask for a explicit question of reboot instead, that works both on automatic and manual install I think, but again the lack of a well defined set of features brought us here and should not guide us further expanding this as to not pile up on tech debt.

mudler: I know that this code is really hairy to say at least, and needs a total refactor... but on the other hand by applying the patch as-is boils down to the fact that then reboot is forced and can't be changed, maybe indeed a NoReboot is better, as reboot is implicitly true by default

Itxaka: >  reboot is forced and can't be changed,

Well, thats automatic install for you :P

mudler (approved):

mudler:
now.. why integration tests didn't triggered... :open_mouth:

PR #48, opened 2022-03-10
Do not require a container image or isoURL

mudler:
During automated installation it was previously required to specify
either a containerImage or a isoUrl, but this is something we impose
ourselves, as elemental otherwise would install from the booting source.

Fixes: https://github.com/rancher-sandbox/cOS-toolkit/issues/1166

### DIFF
--- a/pkg/config/read.go
+++ b/pkg/config/read.go
@@ -245,10 +245,6 @@ func updateData(ctx context.Context, data map[string]interface{}) (map[string]in
 					if newContainerImage == "" {
 						values.PutValue(newData, ContainerImage, "rancheros", "install", "containerImage")
 					}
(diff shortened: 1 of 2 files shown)


Itxaka (approved):
well, that was simple to implement LOL

mudler:
> well, that was simple to implement LOL

Indeed :D I thought it more.. complex, but I think after your refactoring with the containerImage work became more easy too. I'm just annoyed that I'd like to test this code area.. in a way or another.. but still not having good Ideas here which doesn't involve multi machine tests.

I was thinking at whitebox testing.. but then I don't think it's going to add much value besides the config tests we already have, as the interesting part is to be able to see if the installation goes as supposed to..

Issue #49, opened 2022-03-10
luet-mtree is enabled while installing from container images


ldevulder:
Version used: https://github.com/rancher-sandbox/os2/releases/tag/v0.1.0-alpha20

I tried to provision of server with this config:
```
#cloud-config
rancheros:
    tpm:
        emulated: true
        seed: 5
    install:
        registrationCaCert: |-
            -----BEGIN CERTIFICATE-----
            [...]
            -----END CERTIFICATE-----
        registrationURL: https://epinio-k3s/v1-rancheros/registration/vzrn69vhqdjw6w54xrsqdwzkj5rkz2jhmqq8j8vd577x8k72mnthbr
        containerImage: quay.io/costoolkit/os2:v0.1.0-alpha20-amd64
```

Download of the container image was fine, but I ended up with a `luet` error:
```
Mar 10 15:41:22 localhost ros-installer[1756]:   ERROR     Plugin luet-mtree at /usr/bin/luet-mtree had an error: error while executing plugin: exit status 1
Mar 10 15:41:22 localhost ros-installer[1756]: panic: fatal error
```

I attached the whole log file.
[ros-installer.log](https://github.com/rancher-sandbox/os2/files/8224927/ros-installer.log)

kkaempf:
Hmm, a crash in `logger` 🤔 

Where's `luet-mtree` coming from ? rpm package (OBS) ?

ldevulder:
I don't know, I used the ISO image from here: https://github.com/rancher-sandbox/os2/releases/tag/v0.1.0-alpha20

kkaempf:
Ah, these images don't use anything from OBS afaik.

`luet-mtree` is coming from https://github.com/rancher-sandbox/luet-mtree and hasn't seen an update since many weeks.

mudler:
Indeed we should not call mtree here, I think it surfaced as was default in elemental-cli

PR #50, opened 2022-03-11
Disable verify

mudler:
This PR disables `VERIFY` so mtree is not required (see #49).

Besides, tries to add a small test to make sure container image installation works, alongside it also rework the test to use `ros-installer` instead of `elemental install`.


Itxaka (approved):
I was wondering why the provo-mirror appeared there... :D

PR #53, opened 2022-03-16
Drop cos-* references

mudler:
~~Still waiting for https://github.com/rancher-sandbox/cOS-toolkit/pull/1179 to bump cOS~~


ros-image-build
```diff
@@ -260,7 +260,7 @@ VERSION=${IMAGE##*:}
 GIT_COMMIT=${GIT_COMMIT:-HEAD}
 NAME=${IMAGE%%:${VERSION}}
 NAME=${NAME//[^a-zA-Z0-9-@.\/_]/-}
-COS_VERSION=${COS_VERSION:-781cc657aa5ea16b8fe978b8d4bcd19740943db0}
```
Itxaka, line 263: oh my, just had a look at this script....I want to tear my eyes out. why wont we use the containers for all this? Why not have a proper Dockerfile with extra files to build this instead of having everything in there surrounded by quotes? why WHY WHYYYYY

Itxaka (approved):

ros-image-build
mudler: yeap.. I didn't touched anything here because I'm more confident to replace this when elemental have full capabilities to build raw images too and be on pair with this. In that way all of this can just disappear in favor of a much simpler approach (still container based, to avoid deps on the host). 

Also the framework images won't be needed as we already build them in cOS, but that is waiting again for a proper place where to push them (an official registry, or some sorta of) and outside of this card

PR #56, opened 2022-03-18
CI: bootstrapping nodes with MachineRegistration

ldevulder:
Should fix https://github.com/rancher-sandbox/os2/issues/41.


scripts/add-node-in-cluster
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
```
mudler, line 1: nit: I'd keep all the test scenarios inside tests/, maybe we can have a script folder in there too

Just a spare thought, have you ever played with https://github.com/kward/shunit2 ? maybe we can simplify some of e2e tests with that or either bats

.github/workflows/install-with-rancher.yaml
```diff
@@ -0,0 +1,81 @@
...
+      - name: Install dependencies
+        run: zypper -n in -l qemu-kvm libvirt virt-install sudo curl helm git-core tar
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Download artifacts
+        uses: actions/download-artifact@v3
+        with:
+          name: artifacts
+      - name: Install Rancher
+        run: scripts/install-rancher
```
mudler, line 66: maybe we can improve that part and run rancher on either `k3d` or `kind`. In that way we can maybe also run the e2e tests directly in github action runners as well?

.github/workflows/install-with-rancher.yaml
```diff
@@ -0,0 +1,81 @@
...
+      - name: Configure Rancher to provision RancherOS on a test server
+        run: scripts/configure-rancher-step-01
+      - name: Install RancherOS on the test server
+        run: script -e -c scripts/install-ros-node
+      - name: Configure Rancher to add the test server in a test cluster
+        run: scripts/configure-rancher-step-02
+      - name: Add the test server in the test cluster
+        run: scripts/add-node-in-cluster
```
mudler, line 74: The more I look at it as this looks like our first e2e test! can we do that more explicit and group this somewhere (makefile, a common bash entrypoint, ...) that runs it programmatically?

scripts/add-node-in-cluster
ldevulder: > nit: I'd keep all the test scenarios inside tests/, maybe we can have a script folder in there too. This looks like an e2e test, so I'd go ahead creating a specific folder so we can add more scenarios as we go.

I can do this, no problem.

> Just a spare thought, have you ever played with https://github.com/kward/shunit2 ? maybe we can simplify some of e2e tests with that or either bats

Just had a quick look (I already had a look some years ago) but according to the readme it's more to test a shell script "If you have ever had the desire to write a unit test for a shell script, shUnit2 can do the job.". I don't see how it could simplify the test here TBH. The bootstrapping through Rancher needs to be tested, not the shell script itself.

.github/workflows/install-with-rancher.yaml
ldevulder: AFAIK Rancher on K3d is not supported. And moreover if K3d/Rancher is run on a GHA runner I'm afraid that it would not be able to contact a self-hosted/cirrus-ci runner. Rancher can be directly run as a container, I first thinking to use this, but for the reason explained before I don't think that it could run.

scripts/add-node-in-cluster
mudler: > > Just a spare thought, have you ever played with https://github.com/kward/shunit2 ? maybe we can simplify some of e2e tests with that or either bats
> 
> Just had a quick look (I already had a look some years ago) but according to the readme it's more to test a shell script "If you have ever had the desire to write a unit test for a shell script, shUnit2 can do the job.". I don't see how it could simplify the test here TBH. The bootstrapping through Rancher needs to be tested, not the shell script itself.

It can used like bats to organize more complex scenarios, I've been using it successfully with luet, see e.g. https://github.com/mudler/luet/blob/master/tests/integration/01_simple.sh 
but that's outside of the PR scope, it's fine for now to group it somewhere

.github/workflows/install-with-rancher.yaml
ldevulder: I can group this yes, but in that case I will have less steps in the CI and this could be more complicated to debug in case of issue. Or I can split in different steps in a Makefile and add a step to run all at once it needed? I will check this.

.github/workflows/install-with-rancher.yaml
mudler: I was thinking that if we get ahead with k3d or kind (seems kind works on macOS too) and the vm directly on the host, at that point a self-hosted is not even needed because we could run it on vbox in macOS runner. but I'm not sure if kind/rancher works there

.github/workflows/install-with-rancher.yaml
ldevulder: So I can either keep the same kind of scheduling like this:
```
      - name: Install Rancher
        run: cd tests && make install-rancher
      - name: Configure Rancher to provision RancherOS on a test server
        run: cd tests && make configure-rancher-step-01
      - name: Install RancherOS on the test server
        run: cd tests && make install-ros-node
      - name: Configure Rancher to add the test server in a test cluster
        run: cd tests && make configure-rancher-step-02
      - name: Add the test server in the test cluster
        run: cd tests && make add-node-in-cluster
```
Or replace the whole stack by something like this:
```
      - name: e2e - bootstrapping node with Rancher
        run: cd tests && make e2e-bootstraping-node-with-rancher
```

What do you prefer?

.github/workflows/install-with-rancher.yaml
ldevulder: The main issue for me to use macOS is that for now there is x86 macOS runner on GH, but for how long? As they are aarch64 based now ;-)

scripts/add-node-in-cluster
mudler: that would be awesome!

.github/workflows/install-with-rancher.yaml
ldevulder: Test with scripts moved in `tests/` and added in the `Makefile`: https://github.com/ldevulder/os2/actions/runs/2014938295

mudler (approved):
Looks good!

ldevulder:
Merge but still few things to do:
- configure a runner to run it, nested virt on a self-hosted runner? cirrus-ci (a test is in progress on it)?
- schedule it
- convert most of the logic/scripts in Ginkgo

PR #60, opened 2022-03-21
Add flag for ejecting the cd when restarting

Itxaka:
Leverage the elemental CLI flag that allows to eject the cd when
restarting

This is part 3 of https://github.com/rancher-sandbox/os2/issues/31
Requires: https://github.com/rancher-sandbox/cOS-toolkit/pull/1191
Requires: https://github.com/rancher-sandbox/elemental/pull/156

### DIFF
--- a/pkg/install/install.go
+++ b/pkg/install/install.go
@@ -28,12 +28,15 @@ import (
 	"sigs.k8s.io/yaml"
 )
 
-func Run(automatic bool, configFile string, powerOff bool, reboot bool, noRebootAutomatic bool, silent bool) error {
+func Run(automatic bool, configFile string, powerOff bool, reboot bool, noRebootAutomatic bool, silent bool, ejectCd bool) error {
 	cfg, err := config.ReadConfig(context.Background(), configFile, automatic)
 	if err != nil {
 		return err
 	}
 
+	if ejectCd {
+		cfg.RancherOS.Install.EjectCd = true
+	}
 	if powerOff {
 		cfg.RancherOS.Install.PowerOff = true
 	}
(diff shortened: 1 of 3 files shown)


mudler:
This would also need the service to be enabled on boot, or I'm missing something?

Itxaka:
yep, but I think that should be done on the cos-setup part. Im still trying to test this to see if it actually works  :D

mudler:
> yep, but I think that should be done on the cos-setup part. Im still trying to test this to see if it actually works :D

I thought at it as we would just enable it from cloud-config here from the derivatives, as this is optional from cos (It's not needed to have a functional cOS system at the end)

Itxaka:
I mean, we don't lose anything from having it on by default and derivatives can trigger it by just that file creation, I think it's easier to have it as a "feature" of the base os

pkg/install/install.go
```diff
@@ -28,12 +28,15 @@ import (
 	"sigs.k8s.io/yaml"
 )
 
-func Run(automatic bool, configFile string, powerOff bool, reboot bool, noRebootAutomatic bool, silent bool) error {
+func Run(automatic bool, configFile string, powerOff bool, reboot bool, noRebootAutomatic bool, silent bool, ejectCd bool) error {
```
mudler, line 31:
```suggestion
func Run(automatic bool, configFile string, powerOff bool, reboot bool, noRebootAutomatic bool, silent bool, ejectCD bool) error {
```

pkg/install/install.go
```diff
@@ -28,12 +28,15 @@ import (
...
 	cfg, err := config.ReadConfig(context.Background(), configFile, automatic)
 	if err != nil {
 		return err
 	}
 
+	if ejectCd {
```
mudler, line 37:
```suggestion
	if ejectCD {
```

mudler (approved):

PR #68, opened 2022-03-25
Drop rancheros-operator from os2 repo

Itxaka:
It is now living on https://github.com/rancher-sandbox/rancheros-operator
Drops the chart as well.
Modifies all scripts so they go the proper way of installing a chart,
i.e. adding the repo and installing from the repo directly.
Modifies dockerfile to still get the chart into the image as before.


docs/operator.md
```diff
@@ -1,168 +0,0 @@
...
-The RancherOS operator is responsible for managing the RancherOS versions
-and maintaining a machine inventory to assist with edge or baremetal installations.
-
-## Installation
-
-The RancherOS operator can be added to a cluster running Rancher Multi Cluster
-Management server.  It is a helm chart and can be installed as follows:
-
-```bash
-helm -n cattle-rancheros-operator-system install --create-namespace rancheros-operator https://github.com/rancher/os2/releases/download/v0.1.0-alpha12/rancheros-operator-0.1.0-alpha12-amd64.tgz
```
mudler, line 12: That's our only front facing docs available. I would replace it for now with instructions on how to install the chart from the new location, or point to the documentation of the chart if there is any

tests/smoke/smoke_test.go
```diff
@@ -185,13 +178,15 @@ var _ = Describe("os2 Smoke tests", func() {
 				)
 
 				Eventually(func() string {
-					out, _ := s.Command("KUBECONFIG=/etc/rancher/k3s/k3s.yaml helm -n cattle-rancheros-operator-system install --create-namespace rancheros-operator /usr/local/ros.tgz")
+					_, _ = s.Command("KUBECONFIG=/etc/rancher/k3s/k3s.yaml helm repo add rancheros https://rancher-sandbox.github.io/rancheros-operator/")
```
mudler, line 181: My guess here would be to drop this test part entirely (setting up k3s and the operator chart) as we would have separated e2e tests in the chart repo, no? at the end here was about having a small test to see if the ros chart is functional. 

I'd just keep around the k3s part to see if a cluster is functional, without need to test our chart

docs/operator.md
Itxaka: I will link to the ros-operator docs which are on their own repo :)

tests/smoke/smoke_test.go
Itxaka: yes, I just want to see if this work with the os2 iso or I broke something. I guess this is the best chance I got to see if something is missing or broken :+1:

PR #74, opened 2022-03-28
Allow configuring chart url/version

### LINKED ISSUE #70: Allow configuring ros-operator chart url/version
Currently our scripts/os2 binaries use a hardcoded chart url and use the latest DEVEL version (partly due us not releasing a prod version of the ros-operator of course)

There should be some kind of variable that allow us to override both the CHART url repo and the CHART version so we can test with different chart versions/repos.

A good entrypoint for this would be the dapper config which passes configurations into the build process.

One of the problems might be that on some scripts (https://github.com/rancher-sandbox/os2/blob/master/framework/files/usr/sbin/ros-operator-install) the chart is fixed and we cna change that from the build process directly, in fact it uses a locally stored chart in the iso.

So we may require extra configs under /usr/lib/rancheros-release (https://github.com/rancher-sandbox/os2/blob/master/Dockerfile#L161) which IMO is a bad thing. Like the chart version/repo should not be tied to the os version, it should be independent allowing you to upgrade the chart and leave the system running. Before the separation this was not possible, you had the operator tied to the OS, and now its still not possible as the scripts have a hardcoded local path for the operator path and the operator is copied on build time.

So this migth need a rethinking of the ros-operator-install script.

Plus Im not still clear if the ros-operator is an integral part of the os2 feature set or not, as the operator can run anywhere, its not really tied to os2, but seems like os2 wants to provide the feature of having the operator out-of-the-box after install....very weird.

## Requirements

 - [ ] ros-operator-install scripts can get a different url/repo or local file to install
 - [ ] CI scripts (https://github.com/rancher-sandbox/os2/blob/master/tests/scripts/configure-rancher-step-01) can get a different url/repo or local file to install

Itxaka:
fixes #70 

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,17 +4,16 @@ RUN zypper ref
 
 # This target downloads the rancheros operator and makes it available to the framework target
 FROM alpine/helm:3.8.1 as helm
-RUN helm repo add rancheros https://rancher-sandbox.github.io/rancheros-operator
+ARG CHART_REPO=https://rancher-sandbox.github.io/rancheros-operator
+# ">0.0.0-0" means latest including pre-release versions
+ARG CHART_VERSION=">0.0.0-0"
+RUN helm repo add rancheros $CHART_REPO
 RUN mkdir /usr/chart
-# use --devel so alpha versions show up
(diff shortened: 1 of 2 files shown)


mudler (approved):
Looks good here

PR #75, opened 2022-03-28
Do not upload iso/qcow/box with the artifacts - do not download all artifacts for test

Itxaka:
It makes no sense as we upload them twice and we only need the iso/box
in the following jobs


Itxaka:
Saves around 4 minutes per run of integration tests

PR #76, opened 2022-03-28
Try to use the GHA cache when building the images

Itxaka:
Im not sure if this will work, counting that we are running the docker command from inside dapper....

### DIFF
--- a/.github/workflows/integration-tests.yaml
+++ b/.github/workflows/integration-tests.yaml
@@ -12,7 +12,6 @@ env:
    REPO: quay.io/costoolkit/os2-ci
 jobs:
   build:
-
     runs-on: ubuntu-latest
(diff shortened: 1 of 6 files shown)


Itxaka:
Not really possible with our current approach.

 - Running with dapper means that we dont get access to the builder that we set up in the job
 - Running a Dockerfile that references a local image is not supported, because the builder is a docker-container so it doesnt share the images. So building the iso without pushing the image is not possible. This does not affect us as we push the image on the CI but still...
 - No idea why but I cant use the cache properly. Tried pushing to GHA, it did push but didnt use it afterwards, tried local files to /tmp, it stored it but never reused it.

This is something to do down the line I think, once we kind of fix the installer and crap like that, its currently not possible without heavy hacks.

PR #77, opened 2022-03-28
Schedule bootstrapping test

ldevulder:
Only when a PR is merged on `master` branch.

I also added my GH runner with label `nested-virt` to be able to run this test as a workaround, until being able to fix my cirrus-ci run.

### DIFF
--- a/.github/workflows/install-with-rancher.yaml
+++ b/.github/workflows/install-with-rancher.yaml
@@ -34,7 +34,10 @@ jobs:
         uses: actions/upload-artifact@v3
         with:
           name: artifacts
-          path: dist
+          path: |
+            dist/artifacts/*
+            !dist/artifacts/*.qcow.gz
+            !dist/artifacts/*.box
           if-no-files-found: error
       - name: Release space from worker ♻
         if: always()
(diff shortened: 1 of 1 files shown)


.github/workflows/install-with-rancher.yaml
```diff
@@ -1,6 +1,11 @@
-name: Install ROS on a server through Rancher
+name: Bootstrap RancherOS on a VM with Rancher
 
 on:
+  on:
+  pull_request:
```
Itxaka, line 5: why not on branches: master so its only triggered when something is merged onto master

ldevulder: Yes could be, that was mainly to avoid executing it if a direct push is done to `master`. But no problem to change :)

We can also trigger this each night in addition or instead of push on `master` branch, don't know exactly what you prefer guys ;-)

Itxaka: I dont think there is a need to have it on PR closed, just when something is merged to master. There should not be anyone merging directly to master without opening a PR first :)

Itxaka: And that means you can remove the if below and having as a straight job

Itxaka (approved):

.github/workflows/install-with-rancher.yaml
ldevulder: OK, so I changed to schedule the test when a push is done on `master`, it should be fine now.

Itxaka (approved):

PR #81, opened 2022-03-30
ci: fix missing env variables

ldevulder:


mudler (approved):

.github/workflows/install-with-rancher.yaml
```diff
@@ -7,27 +7,15 @@ on:
...
 
 jobs:
   build:
     runs-on: ubuntu-latest
     env:
       DOCKER_USERNAME: ${{ secrets.QUAY_USERNAME }}
       DOCKER_PASSWORD: ${{ secrets.QUAY_PASSWORD }}
-      DOCKER_REGISTRY: ${{ github.event.inputs.registry }}
-      REPO: ${{ github.event.inputs.registry }}/${{ github.event.inputs.org }}/os2-ci
+      DOCKER_REGISTRY: quay.io
```
Itxaka, line 17: arent you able to get the default inputs by getting the env `INPUT_NAME` like `INPUT_REGISTRY` ? I seem to rememeber that being a possibility, let me check if I can find that

Itxaka: mmmh, nope that seems to be only available to actions, sorry :/

ldevulder: Yes, I tried to avoid fixed values but I wasn't able to reuse the default ones... ANother solution could be to use secrets, but it's not so good to not been able to see these kind of non-private values IMO.

PR #89, opened 2022-04-08
Convert bootstrapping test scripts in go

### LINKED ISSUE #62: Convert bootstrapping-with-rancher test to Ginkgo
As most of the current tests are using Ginkgo, could be easier to have the new ones also executed with Ginkgo framework.

ldevulder:
Should fix #62.

**NOTE:** for this version, as a TTY is needed for `virt-install` I still use a shell script called with `script` command. I will do some tests to replace this purely in GO, but more for another version. Same for the HTTP server, I have some Go code but it's not working as expected, I will try to fix this later.


tests/e2e/helpers/tools/tools.go
```diff
@@ -0,0 +1,97 @@
...
+	if err != nil {
+		return err
+	}
+
+	// Save data in file
+	_, err = io.Copy(file, data.Body)
+	return err
+}
+
+func GetFiles(dir string, pattern string) ([]string, error) {
```
mudler, line 52: this file looks a good candidate that could go into `ele-testhelpers`

mudler (approved):
Looks good! really nice!

tests/e2e/helpers/tools/tools.go
ldevulder: Yes, can I do it in another PR? As I will have to do 2 PR in 2 differents repos.

mudler: yup definetly, this is just a suggestion, makes completely sense having a follow-up for it

ldevulder: OK, I will do it quickly.

ldevulder: Done here: https://github.com/rancher-sandbox/os2/pull/102

PR #91, opened 2022-04-12
ci: use new runner for bootstrap test + force Golang v1.17.x

### LINKED ISSUE #61: Add/configure/find-a-way to use virtualization for VM tests
Current tests on VM run on MacOS runner in GH with VirtualBox because virtualization is not available on Linux worker. MacOS runners are x86 for now but will be moved to aarch64 at some point in the future. So could be a good idea to find another way to run these tests.

Rancher-Desktop uses [cirrus-ci](https://cirrus-ci.org) for this, as KVM can be configured on container image.

- [x] configure a runner to run it, nested virt on a self-hosted runner? cirrus-ci (a test is in progress on it)?
- [x] schedule test(s) that need virtualization (new, not the ones that already use vbox)

ldevulder:
Should fix #61.

This PR also force Golang v1.17.x

### DIFF
--- a/.github/workflows/install-with-rancher.yaml
+++ b/.github/workflows/install-with-rancher.yaml
@@ -10,7 +10,7 @@ on:
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    runs-on: kvm-host
     env:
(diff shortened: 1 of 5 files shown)


mudler (approved):
Looks good!

PR #93, opened 2022-04-13
ci: use build-host for build process

ldevulder:
This should speedup the build processes.

E2E verification runs:
- latest one! https://github.com/rancher/elemental/actions/runs/2651932206
- ~with standard selfhosted runner conf: https://github.com/rancher/elemental/actions/runs/2529263149/attempts/1~
- ~with `_work` on tmpfs and `tuned` configured for performance: https://github.com/rancher/elemental/actions/runs/2529263149/attempts/2 (it seems that the I/O is degraded, but why?)~
- ~with `tuned` configured for performance: https://github.com/rancher/elemental/actions/runs/2529263149/attempts/3~


mudler (approved):
Looks good!

ldevulder:
The build issue is due to the fact that all runs directly on the runner and not on an container image (Ubuntu in that case), on it to fix this.

ldevulder:
I still have this error:
`Error while creating domain: Error saving the server: Call to virDomainDefineXML failed: End of file while reading data: Input/output error`

Even after forcing the build step to use an Ubuntu image...

This PR was to reduce build time by using a self hosted runner but I really don't know how to workaround this error... So better to close it for now I think.

ldevulder:
I finally found why `tests` is failing!

Here a working test:
```
==> os2: Creating image (snapshot of base box volume).
==> os2: Creating domain with the following settings...
==> os2:  -- Name:              tests_os2
==> os2:  -- Description:       Source: /home/runner/work/elemental/elemental/tests/Vagrantfile
==> os2:  -- Domain type:       qemu
==> os2:  -- Cpus:              4
```

And here the failing one:
```
==> os2: Creating image (snapshot of base box volume).
==> os2: Creating domain with the following settings...
==> os2:  -- Name:              tests_os2
==> os2:  -- Description:       Source: /home/runner/work/elemental/elemental/tests/Vagrantfile
==> os2:  -- Domain type:       kvm
==> os2:  -- Cpus:              4
```

The domain type is `kvm` when compiled on the self-hosted runner! I think this is because nested virtualization is on. I will try to see if this could be forced during building process.

ldevulder:
Ok so I think I understand the process and why KVM accel is set: it's detected when `ros-image-build` is executed [(see here](https://github.com/rancher/elemental/blob/4c180c50a8f197d04a886ecf2859b425f5a6a08a/ros-image-build#L160)).

Which lead me to ask: **why is this so complicated??**
I checked a bit and found that the GH workflow execute `make ci` which execute `dapper ci` which uses `Dockerfile.dapper` to generate and run a docker image that execute `scripts/ci` which itself execute `scripts/package` which itself execute `make build` and `make iso` (and also others `make ...` commands). This `build` command generates a docker image that is used in `make iso` and this command execute `ros-image-build` that generate docker image and the other artifacts (iso, qcow2, box, etc.) by also using some specific tools like `packer` (which is useful for some cases I agree).

But this is really difficult to understand TBH...

mjura (approved):
LGTM

Itxaka (approved):
:shipit:

kkaempf:
@mjura @ldevulder - what is this waiting for ?

ldevulder:
> @mjura @ldevulder - what is this waiting for ?

A verification run that works... I have issue and I'm currently doing manual tests to understand why (could be due to some renaming change). If I can fix in this PR I will do, otherwise I will merge this one and create new to fix.

PR #95, opened 2022-04-20
Generate default upgrade channel image file in releases

mudler:
Part of https://github.com/rancher-sandbox/os2/issues/87

~Missing docs updates~ Added


scripts/package
```diff
@@ -60,6 +60,33 @@ initrd \${url}/\${initrd}
 boot
 EOF
 
+GITHUB_REPOSITORY="${GITHUB_REPOSITORY:-rancher-sandbox/os2}"
+UPGRADECHANNEL_DISCOVERY_IMAGE="${UPGRADECHANNEL_DISCOVERY_IMAGE:-quay.io/costoolkit/upgradechannel-discovery:v0.3-4b83dbe}"
```
mudler, line 64: Not sure where to put that honestly. I'd move them up to the Dockerfile, but I noticed in this file there are already other variables set inline, so we should probably go back at this and refactor this script properly separately

scripts/package
```diff
@@ -60,6 +60,33 @@ initrd \${url}/\${initrd}
...
+cat > build/output.upgradechannel-$i.yaml << EOF
+apiVersion: rancheros.cattle.io/v1
+kind: ManagedOSVersionChannel
+metadata:
+  name: os2-$i
+  namespace: fleet-default
+spec:
+  options:
+    args:
+    - github
```
mudler, line 75: I opted out for scanning just github releases here with rancher-sandbox/os2. Both for sake of semplicity on our side and flexibility on user side. As this is for the community one, we should be good with it to keep synced to always latest available versions.

For other more convoluted scenarios (which might involve obs, et all) we can use the `git` syncer instead so we are in control of what versions are going to be pushed

Itxaka (approved):

mudler:
This is an example yaml file which was generated by the CI: 

```yaml
apiVersion: rancheros.cattle.io/v1
kind: ManagedOSVersionChannel
metadata:
  name: os2-amd64
  namespace: fleet-default
spec:
  options:
    args:
    - github
    command:
    - /usr/bin/upgradechannel-discovery
    envs:
    - name: REPOSITORY
      value: rancher-sandbox/os2
    - name: IMAGE_PREFIX
      value: quay.io/costoolkit/os2-ci
    - name: VERSION_SUFFIX
      value: -amd64
    image: quay.io/costoolkit/upgradechannel-discovery:v0.3-4b83dbe
  type: custom
```

Note, for the user is enough to apply it with a cluster with `rancheros-operator` being installed. The namespace here is set by default to `fleet-default`, but not that can be changed to `fleet-local` to target instead the local cluster.

mudler:
Docs updated

mudler:
Dunno why, but getting all linting issues now also for code that was previously there :sweat_smile: fixing them..

PR #96, opened 2022-04-25
Use elemental's build-iso command to build the ISO

davidcassany:

### DIFF
--- a/ros-image-build
+++ b/ros-image-build
@@ -59,45 +59,20 @@ RUN echo -e \
 '  urls:\n'\
 '  - "quay.io/costoolkit/releases-green-arm64"\n' > /etc/luet/luet.yaml && \
 echo -e \
-'packages:\n'\
+'iso:\n'\
+'  rootfs:\n'\
+'  - /iso/overlay\n'\
 '  uefi:\n'\
 '  - live/grub2-efi-image\n'\
-'  isoimage:\n'\
+'  image:\n'\
 '  - live/grub2\n'\
 '  - live/grub2-efi-image\n'\
+'  - /iso/iso-overlay\n'\
+'  label: "COS_LIVE"\n'\
 '\n'\
-'boot_file: "boot/x86_64/loader/eltorito.img"\n'\
-'boot_catalog: "boot/x86_64/boot.catalog"\n'\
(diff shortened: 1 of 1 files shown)


Itxaka (approved):

mudler:
Do we still need to build a config file?

davidcassany:
> Do we still need to build a config file?

You mean the manifest embedded inside the Dockerfile? if so yes, there are no defaults set for the included packages as there is no Luet default configuration too. Note that those packages relay on  a Luet repository, so I am not sure having a default for everything is that good (I mean packages list and luet repositories). Anyway this is something simple to add if we want, it is all configurable via manifest and flags, so adding some extra defaults is trivial.

mudler:
> > Do we still need to build a config file?
> 
> You mean the manifest embedded inside the Dockerfile? if so yes, there are no defaults set for the included packages as there is no Luet default configuration too. Note that those packages relay on a Luet repository, so I am not sure having a default for everything is that good (I mean packages list and luet repositories). Anyway this is something simple to add if we want, it is all configurable via manifest and flags, so adding some extra defaults is trivial.

I thought was already like this so we could get rid of the base file as there are no big customization here (just the iso overlay, which I thought you could pass with some arg), at the end a full manifest could be still passed by in case of overrides. Nevertheless looks good anyway, it's a nice cutdown!

mudler (approved):

davidcassany:
> I thought was already like this so we could get rid of the base file as there are no big customization here (just the iso overlay, which I thought you could pass with some arg), at the end a full manifest could be still passed by in case of overrides. Nevertheless looks good anyway, it's a nice cutdown!

I did not add these defaults for everything as I had in mind the idea of having it always executed on top of a generic manifest and apply all little customizations on top of it. I also thought about having a default `--config-dir` value set to `.` or something like so we load the manifest from the current folder (even this does not really apply on this `os2` setup, but nice from a cOS-Toolkit PoV). Anyway the main reason to not add thiese defaults is having to hardcode the luet repository URL. To me it feels odd hardcoding it inside elemental, but we can easily do a follow up PR for that and discuss it, I do not have any technical reason against it.

PR #99, opened 2022-04-27
Remove manifest.yaml file to build ISO and relay on defaults

davidcassany:

### DIFF
--- a/ros-image-build
+++ b/ros-image-build
@@ -57,21 +57,7 @@ RUN echo -e \
 '  enable: true\n'\
 '  arch: arm64\n'\
 '  urls:\n'\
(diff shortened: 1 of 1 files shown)


mudler (approved):
Loving it!

davidcassany:
Previously failed due to a missing default label in elemental. Added the label flag as a workaround to verify it works, but the fix should com from elemental. The label flag should not be needed at all. That's why I targeted it as draft, the required fix should be part of elemental.

Itxaka (approved):

davidcassany:
The fix from elemental is coming here rancher-sandbox/elemental#179

PR #112, opened 2022-05-09
ci: use scsi device instead of virtio for e2e tests

ldevulder:
This commit also fix the install device issue (should be under `rancheros` section).
Also use a "real" password for root user.


kkaempf:
This still doesn't create (resp. enable) the root account for me.

mudler (approved):

PR #117, opened 2022-05-11
'ros-installer' runs automatically

kkaempf:

### DIFF
--- a/framework/files/system/oem/05_rancheros-installer.yaml
+++ b/framework/files/system/oem/05_rancheros-installer.yaml
@@ -6,7 +6,8 @@ stages:
     - path: /etc/motd
       content: |
 
-        Run "ros-installer" to install to disk
+        "ros-installer" is already running.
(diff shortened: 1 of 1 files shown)


framework/files/system/oem/05_rancheros-installer.yaml
```diff
@@ -6,7 +6,8 @@ stages:
     - path: /etc/motd
       content: |
 
-        Run "ros-installer" to install to disk
+        "ros-installer" is already running.
```
mudler, line 9: this might not be always the case. this would be true if there is a config file for automated installs

kkaempf: Right. But we're fixing the 05_**rancheros**-installer.yaml here :wink:

mudler: sure, but this file runs everytime in `live_mode`. it does not take into account what's inside the user configuration.

To do it properly should be gated with the if above somehow.

PR #122, opened 2022-05-13
Add node upgrade e2e test

ldevulder:
Should fix https://github.com/rancher-sandbox/os2/issues/42 and https://github.com/rancher-sandbox/cOS-toolkit/issues/1280

TODO:
- [x] Split steps  in multiple files
- [x] Create functions to avoid duplicate code
- [x] Create a function to configure iPXE script depending on the version to install
- [x] Adapt `GetServerId` function to check the installed VM and not only the first one
- [x] Refactor the code to not duplicate the installation steps
- [x] Add node upgrade with `osImage` test scenario
- [x] Add `upgradechannel` test scenario

Verification run logs: [verification_run.log](https://github.com/rancher/elemental/files/8822437/verification_run.log)

NOTE: the only thing that was not tested in ***real*** conditions is the GHA workflow, I tested it manually steps by steps.


mjura (approved):
This PR looks great

PR #128, opened 2022-05-16
Run tests also on pull requests

mudler:
As we don't have anymore the chart to test here, there is no need to push the CI image anymore during tests. This allows us to also use ttl.sh and also firing up tests on PRs


.github/workflows/integration-tests.yaml
```diff
@@ -1,15 +1,15 @@
...
 concurrency:
   group: integration-tests=full-${{ github.head_ref || github.ref }}-${{ github.repository }}
   cancel-in-progress: true
 
 env:
-   DOCKER_USERNAME: ${{ secrets.QUAY_USERNAME }}
-   DOCKER_PASSWORD: ${{ secrets.QUAY_PASSWORD }}
-   DOCKER_REGISTRY: quay.io
-   REPO: quay.io/costoolkit/os2-ci
+   DOCKER_REGISTRY: ttl.sh
```
ldevulder, line 10: I think I will have to do the same for the e2e test, no?

Itxaka (commented):
This will duplicate any pr that comes from a branch of the original repo. This only works for fork+pr. I dunno, duplicate stuff with the same sha usually leads to issues... You know I'm not a friend of that 😉

.github/workflows/integration-tests.yaml
mudler: they run on master so shouldn't be necessary

mudler:
> This will duplicate any pr that comes from a branch of the original repo. 

You mean that you will run 2 jobs? If there is a nicer way to do with github without going into any complex workflow setups.. I'm all for it.

> This only works for fork+pr. I dunno, duplicate stuff with the same sha usually leads to issues... You know I'm not a friend of that wink

any suggestion?

Itxaka (approved):

mudler:
@Itxaka thanks for reviewing this!

PR #129, opened 2022-05-17
Drop rancherd and ros-installer service files

kkaempf:
Both binaries have their own repos now, where these files belong.


mudler:
mmm tests are failing. Also, I don't see those services in their respective repositories?

kkaempf:
not needed anymore

PR #156, opened 2022-06-22
Add System Agent and remove RancherD

paynejacob:
This pr removes all rancherd and ros-installer systemd unit files and configuration. Cluster provisioning will now be handled by rancher.  Binaries for elemental-operator and system agent were added.  The elemental operator is used to download cloud config to inform the elemental cli install command as well as register the inventory with rancher.  The system agent is used to allow the operator to execute arbitrary commands on the inventory as well as copy installation files.


Dockerfile
```diff
@@ -4,6 +4,8 @@ FROM $BASE_IMAGE
 # Framework files
 COPY framework/cos/ /
 COPY framework/files/ /
+COPY --from=rancher/system-agent:v0.2.7 /usr/bin/rancher-system-agent /usr/sbin/rancher-system-agent
+COPY --from=ghcr.io/paynejacob/elemental-operator:dev /usr/bin/elemental-operator /usr/sbin/elemental-operator
```
mudler, line 8: Maybe a oversight? shouldn't this come from rancher/elemental-operator?

framework/files/usr/lib/systemd/system/elemental-system-agent.service
```diff
@@ -0,0 +1,17 @@
+[Unit]
+Description=ROS System Agent
```
mudler, line 2:
```suggestion
Description=Elemental System Agent
```

Dockerfile
```diff
@@ -4,6 +4,8 @@ FROM $BASE_IMAGE
 # Framework files
 COPY framework/cos/ /
 COPY framework/files/ /
+COPY --from=rancher/system-agent:v0.2.7 /usr/bin/rancher-system-agent /usr/sbin/rancher-system-agent
```
rosskirkpat, line 7: bump to 0.2.8 of SA

Dockerfile
```diff
@@ -4,6 +4,8 @@ FROM $BASE_IMAGE
...
+COPY --from=ghcr.io/rancher/elemental-operator:dev /usr/bin/elemental-operator /usr/sbin/elemental-operator
```
rosskirkpat, line 8: Is elemental using DockerHub or GHCR?

rosskirkpat (changes requested):
LGTM with a request and question

framework/files/etc/systemd/system/elemental-system-agent.service.d/override.conf
```diff
@@ -0,0 +1,6 @@
+[Unit]
+After=
+After=cos-setup-network.service
+
+[Service]
+ExecStartPre=-/usr/bin/mkdir -p /usr/local/bin
```
mjura, line 6: Missing newline character

mjura (commented):
CI is failing

framework/files/usr/lib/systemd/system/elemental-install.service
```diff
@@ -0,0 +1,18 @@
...
+WantedBy=multi-user.target
+
+[Service]
+Type=oneshot
+EnvironmentFile=-/etc/default/%N
+EnvironmentFile=-/etc/sysconfig/%N
+TimeoutStartSec=0
+StandardOutput=journal+console
+StandardError=journal+console
+ExecStartPre=/bin/sh -c 'echo "ELEMENTAL_TARGET=/dev/$(lsblk -r -o NAME,TYPE | grep -w disk | grep -v fd0 | awk \'{print $1}\')" > /etc/sysconfig/elemental-install'
```
agracey, line 16: I'm still nervous about this line. I think the value set here is taking precedence over what set in the MachineRegistration

kkaempf:
None of this is relevant for Elemental Teal.

The respective RPMs bring their own `.service` files and service enablement happens via [systemd-presets-branding-Elemental](https://build.opensuse.org/package/show/isv:Rancher:Elemental/systemd-presets-branding-Elemental)

Dockerfile
mjura: Image can't be pulled `Error response from daemon: Head "https://ghcr.io/v2/rancher/elemental-operator/manifests/dev": denied`

Itxaka: nobody is pushing to that repo so not sure why this is here :D

framework/files/usr/lib/systemd/system/elemental-install.service
fgiudici: My guess: to provide a default if no value is provided by the MachineRegistration. If this is the case, it could make sense because you can try to detect the device you want to install to only when you are on the actual host for the installation.
But the main point is: does the ELEMENTAL_TARGET takes precedence over the yaml config passed to "elemental install" ?

Dockerfile
fgiudici: In the first version it was a privarte repo, it was than changed to this... but as you said, nobody is pushing there.

framework/files/usr/lib/systemd/system/elemental-install.service
Itxaka: We probably need validation then to make sure that at least a target is added OR make elemental-installer use an "auto" stanza and make it do the same as here instead of hardcoding files with something dangerous.


I can't recall, but IIRC env args have precedence over config files. I would need to recheck this thougth

fgiudici: I completely agree with you. This really looks a bit hacky and is unexpected (or even buggy, if it takes precedence)

agracey: This might be a moot point: https://github.com/rancher/elemental/pull/156#issuecomment-1175968783

framework/files/etc/systemd/system/elemental-system-agent.service.d/override.conf
kkaempf: This is covered by https://github.com/rancher/elemental/pull/172

kkaempf:
Closing for now.

PR #167, opened 2022-07-14
Simplify building elemental iso

### LINKED ISSUE #165: elemental-iso-build (replacing ros-image-build ?)
`ros-image-build` is outdated and should be removed.

There's [elemental-iso-build](https://github.com/rancher-sandbox/rancher-node-image/blob/main/elemental-iso-build) which should replace it.
This build script should
- use registry.opensuse.org/isv/rancher/elemental/teal52/15.3/rancher/elemental-node-image/5.2:last as the base
- use registry.opensuse.org/isv/rancher/elemental/teal52/15.3/rancher/elemental-builder-image as "tools"
- only focus on ISO creation (qcow etc can be added later, but we don't need it right now)
- be simple to run (like `elemental-iso-build <registration>.yaml` instead of the current `ros-image-build <image> iso <registration>.yaml`

Current problems with [elemental-iso-build](https://github.com/rancher-sandbox/rancher-node-image/blob/main/elemental-iso-build)
- calls `elemental build-iso` wrongly (can't find rootfs)
- is still based on `ros-installer`, probably still carries some unneeded "cruft"

Itxaka:
Drop the script, use the Makefile directly.
Set the default files used under the iso/ dir
Set proper directories for overlay and rootfs
Simplify the Dockerfile used
Allow specifying a cloud-config file by setting CLOUD_CONFIG_FILE before
building the image, otherwise default to the generic one

Fixes #165 

### DIFF
--- a/.github/workflows/e2e-tests.yaml
+++ b/.github/workflows/e2e-tests.yaml
@@ -30,7 +30,9 @@ jobs:
       - name: Set up Docker Buildx
         id: buildx
         uses: docker/setup-buildx-action@v2
-      - name: Make images (container/iso/qcow2/etc.)
+      - name: Make ISO
+        run: make iso
+      - name: Make images (container/qcow2/etc.)
         run: make ci
       - name: Cache data
         uses: actions/cache@v3
--- a/.github/workflows/e2e-tests.yaml
+++ b/.github/workflows/e2e-tests.yaml
@@ -30,9 +30,7 @@ jobs:
       - name: Set up Docker Buildx
         id: buildx
         uses: docker/setup-buildx-action@v2
-      - name: Make ISO
-        run: make iso
-      - name: Make images (container/qcow2/etc.)
+      - name: Make images (container/iso/qcow2/etc.)
         run: make ci
       - name: Cache data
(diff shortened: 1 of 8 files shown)


kkaempf (approved):
Thanks !

kkaempf:
Be careful with moving too much into `Makefile`
While this is fine for us engineers, others would rather work with a single script instead of a git checkout 😉

Itxaka:
> Be careful with moving too much into Makefile
While this is fine for us engineers, others would rather work with a single script instead of a git checkout wink

Thats their problem. If you know how to run a script, you know how to run make iso, make packer, make whatever. And if they dont, they got all the artifacts in the release prepared to be consumed :)

Otherwise, this is hellish to maintain with a decent amount of care. Like Im pretty sure we are building wrong some of this stuff, but its impossible to understand why or how or where it it comes from. Is unacceptable to release something when you dont have any faith or knowledge of the build chain.

Itxaka:
of course tests fail becuase they are trying to run ros-installer.....which shows that we were bulding the image wrong before? no more ros-installer should be in the images AT ALL.

PR #168, opened 2022-07-14
Add a simple script to build the iso

Itxaka:
So people can use to to build their own isos with embedded cloud-config
files


Itxaka:
no git, no docker, no nothing. you just curl a random internet into bash as see the magic happen :P

kkaempf (approved):
Thank you !!

PR #173, opened 2022-07-18
Fix the motd message

kkaempf:
also rename the file to reflect its purpose


kkaempf:
Local tests passed.

Issue #169, opened 2022-07-15
Rework the way we build artifacts


Itxaka:
Currently we have a mess in the way we build the artifacts for the repo. Some mix of docker in docker, dapper, strange targets, packer inside docker inside dapper, config files embedded in bash scripts and so on.

This makes it very difficult to understand  the pipelines and understand where the artifacts are coming from, what are we building and what is the final product.

We should be able to drop the ros-image-build and all the dapper files and recrete decent targets to build from. I.E. packer templates to build the qcow+box, a Dockerfile to build the container and the iso, which is already in the Makefile.

Then we could create a clean CI pipeline in which every built thing its own job so we can clearly see what is doing and understan when there is failures.

Also, lets stop pushing to ssh.tl for CI as those images are dropped after a few hours. We can create an elemental-ci repo and push the images there as we do with the operator so we can test them and come back.

Also a decent release strategy should be created.

Basically, lets see what we have already on the operator, toolkit, cli and use it in this repo to bring it up to standars with the rest of our projects instead of having a black box that nobody knows about.

ldevulder:
In cas it could help a bit :-), here what I found: https://github.com/rancher/elemental/pull/93#issuecomment-1156292431

Itxaka:
gonna claim this as done, there is card for further steps

PR #174, opened 2022-07-18
Drop ros image build

kkaempf:
Drop obsolete `ros-image-build`


Itxaka:
This will break all tests. https://github.com/rancher/elemental/issues/169

ldevulder:
> Drop obsolete `ros-image-build`

But nothing to replace it?

Itxaka:
> We have elemental-iso-build now

Nope, that doesn't work. They are completely different and it affects the whole CI pipeline. You can try substituting it in this PR and you will see :D

 - elemental-iso-build builds the iso FROM released obs images. It makes no sense to use that in the CI pipeline to test the current code as that completely ignores it.
 - elemental-iso-build doesnt bundle ros-installer anymore. e2e tests require that
 - moving from ros-installer to elemental-operator register means that we need a rancher server spin up for e2e tests...kind of a big thing
 - if instead moving from ros-installer to elemental install, we still need stuff to adapt AND that will not properly test the use case that we want (boot a machine with elemental and a registration-url, see that it ends up working)

ldevulder:
> * moving from ros-installer to elemental-operator register means that we need a rancher server spin up for e2e tests...kind of a big thing

It's already done :-) the E2E tests install k3s+rancher and then create 4 VMs for test OS bootstrapping, cluster creation and upgrade.

ldevulder:
These changes have been done (with build refactoring!) in Itxaka PR #178, so this PR could be closed.

PR #178, opened 2022-07-19
Rework image/iso/artifacts building

Itxaka:
Drop all dapper stuff

Build elemental image from teal releases.
Build iso based of elemental image.
Tag images and isos properly
Install and enable elemental-operator and rancher-system-agent
Drop uneeded files
Add config files so we can build embedded configs on iso images
Remove old ros-installer stuff
Run elemental-operator as part of the livecd cloud config
Set the elemental version on the grub install line
Add docker build ci job
Set proper versions all over the place. No more `elemental:dev` !
Use sane defaults for values everywhere.
Build iso based of generated image, not a random image. Allows overriding base image for the iso.
Allow overriding mostly everything so its flexible for building locally on dev and on other places (maybe OBS? dunno)
Clean ipxe files as they were using the old ros-installer cmdline

NOTE: this does not contain the end2end enabled tests or the qcow/box images. Those can come afterwards.

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -2,7 +2,7 @@
 
 Elemental is a toolkit to build an immutable Linux distribution.
 
-It's primary purpose is to run Rancher and its corresponding Kubernetes distributions [RKE2](https://rke2.io) 
+Its primary purpose is to run Rancher and its corresponding Kubernetes distributions [RKE2](https://rke2.io) 
 and [k3s](https://k3s.io).
 But it can be configured for any other workload. That said, the following documentation focusses on a Rancher use-case.
 


ldevulder:
> AFAIK that's the end goal to add nodes to rancher unnatended so that is probably the goal of the end2end tests, have a rancher manager installed and add nodes with a registrationurl and watch them appear?

Yes check the [README](https://github.com/rancher/elemental/blob/master/tests/e2e/README.md) for a full list of the steps done in E2E tests.

Itxaka:
> What do you mean with this? Are you keeping the E2E tests as-is? Even if they fail we will take care of this.

Yep, will restore the job yaml to see it fail, but will need to first adapt how we build the artifacts as that has changed

ldevulder:
> Yep, will restore the job yaml to see it fail, but will need to first adapt how we build the artifacts as that has changed

Ok, so as soon as you fix the artifacts build we will handle the E2E part with @juadk.

Dockerfile.image
```diff
@@ -0,0 +1,68 @@
+# ARGS go first if used on FROM
+ARG OPERATOR_IMAGE=quay.io/costoolkit/elemental-operator:v0.3.0
+ARG SYSTEM_AGENT_IMAGE=rancher/system-agent:v0.2.9
```
ldevulder, line 3: Isn't better to use the `latest` released version instead of a fixed one?

Dockerfile.iso
```diff
@@ -1,10 +1,16 @@
-FROM registry.opensuse.org/isv/rancher/elemental/teal52/15.3/rancher/elemental-node-image/5.2:latest AS os
-FROM registry.opensuse.org/isv/rancher/elemental/teal52/15.3/rancher/elemental-builder-image as tools
+ARG OS_IMAGE=registry.opensuse.org/isv/rancher/elemental/teal52/15.3/rancher/elemental-node-image/5.2:latest
+ARG TOOL_IMAGE=registry.opensuse.org/isv/rancher/elemental/teal52/15.3/rancher/elemental-builder-image
```
ldevulder, line 2: I think that without `:lastest` at the end it's anyway the latest one by default?

Makefile
```diff
@@ -1,119 +1,123 @@
...
-ci: .dapper
-	./.dapper ci
-
-.PHONY: package
-package: .dapper
-	./.dapper package
+# This are the default images already in the dockerfile but we want to be able to override them
+OPERATOR_IMAGE?=quay.io/costoolkit/elemental-operator:v0.3.0
+SYSTEM_AGENT_IMAGE?=rancher/system-agent:v0.2.9
+TOOL_IMAGE?=quay.io/costoolkit/elemental:v0.0.15-f1fabd4
```
ldevulder, line 14: Same comment as above: why not using the latest versions?

framework/files/etc/luet/luet.yaml
```diff
@@ -0,0 +1,16 @@
+repositories:
+  - &cos
+    name: "cos"
+    description: "cOS official"
```
ldevulder, line 4: Shouldn't be `Elemental` instead of `cOS`?

ldevulder (approved):
Looks good, we will now re-add the E2E tests!

Dockerfile.image
Itxaka: Not really, I tried to use the versions that are being used on OBS currently. Maybe once we got tests we can move this into a latest tag or maybe have an extra job that uses the latest tags for everything so we know if something breaks before it's changed in OBS, kind of a staging job

Dockerfile.iso
Itxaka: Yep, but we don't care, this repos are used to build an image with the obs published images and by the get-elemental-iso script that can be used from outside this repo so we want to default to the obs images, the official ones.

Alia I don't think we currently publish tags for those images, there is only one published at a time

Makefile
Itxaka: For this one, I need to fix the publish job for elemental-cli as it seems that it's not pushing the proper version on tag, only the tag+sha yikes

framework/files/etc/luet/luet.yaml
Itxaka: Mmh, this are generic repos based off the teal flavor so they are generic enough that they can be used by any derivativate that has a sle micro base. So not really elemental related but maybe a better name could be used like upstream repos or elemental-toolkit repos

Itxaka:
Can't break tests if you remove tests

ldevulder:
> If it breaks the CI, you fix it

I think he already kinda fix it! :rofl:

Issue #166, opened 2022-07-14
There's still ros-operator-install and ros-populate-node-labels in framework/files/usr/sbin


kkaempf:
Are these still needed ?

Itxaka:
for the e2e testing they are still being used. Either we break the e2e testing by dropping them or we rework the e2e testing to use...elemental-installer? Which means we need to setup a rancher cluster in order to test elemental.

kkaempf:
From discussions on the daily
- drop ros-operator-install
- rename ros-populate-node-labels to elemental-populate-node-labels

PR #180, opened 2022-07-20
Drop ros-operator-install and rename populate-node-labels script

mudler:
Fixes: https://github.com/rancher/elemental/issues/166


kkaempf (approved):
lgtm

kkaempf:
> mmh. ros-operator-install seems to be used here:
> 
> https://github.com/rancher/elemental/blob/c3e62ce672b1f3cf3babd41df9cac9426df84326/framework/files/usr/share/rancher/rancherd/config.yaml#L2
> 
> but as rancherd is no more.. I guess we shouldn't care?

Right.
That's probably worth an issue in rancher/elemental.

Itxaka:
Ros-operator-install was dropped as part of #178

Itxaka:
I think this is still being called somewhere in the systemd services and the systemd services are also names wrongly 🙂

Issue #171, opened 2022-07-15
Properly configure system agent persistent paths


davidcassany:
system agent requires persistence under some paths like `/var/lib/elemental` and/or `/var/lib/rancher`. We need to ensure the `immutable-rootfs` setup is aligned with the system agent persistence requirements. 

[immutable-rootfs docs](https://rancher.github.io/elemental-toolkit/docs/reference/packages/system-immutable-rootfs/)

davidcassany:
> Is that something we need to create in Elemental Teal ?

If Elemental Teal refers to the actual OS image we want to install in nodes I'd say yes. More specific I think we just need to ensure that the persistent paths required by system agent need to be under `/usr/local/` or be listed within the `framework/files/system/oem/01_ros_rootfs.yaml` file as part of the `PERSISTENT_STATE_PATHS` immutable rootfs config variable.

Itxaka:
This is already in place in the PERSISTENT_STATE_PATHS of elemental..closing @davidcassany ?

davidcassany:
Yes I think this can be closed for now as the system works. A couple of weeks ago it was unclear which paths should be added here, but I guess we are just fine for the time being.

PR #181, opened 2022-07-20
Make /var/lib/elemental persistent

kkaempf:
This esp. allows proper tracking of local plans

Part of #171


mudler (approved):
Looks good here. Also the path should be already present on the rootfs, but i think after @davidcassany recent changes this is not needed anymore. Can you please confirm?

davidcassany (approved):

PR #187, opened 2022-07-21
Renames populate node labels services

Itxaka:
bin was renamed but not the services

Signed-off-by: Itxaka <igarcia@suse.com>

PR #188, opened 2022-07-21
Cleanup RancherOS references

mudler:

### DIFF
--- a/.github/workflows/build-iso.yaml
+++ b/.github/workflows/build-iso.yaml
@@ -1,4 +1,4 @@
-name: Build iso from simple script
+name: Test build iso from simple script
 
 on:
   push:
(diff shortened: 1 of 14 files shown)


mudler:
Supersedes https://github.com/rancher/elemental/pull/187

framework/files/usr/sbin/elemental-populate-node-labels
```diff
@@ -6,10 +6,10 @@ set -e
 IMAGE=$(echo "$IMAGE" | sed 's![/:]!-!g')
 for i in rke2 k3s; do
     mkdir -p /etc/rancher/$i/config.yaml.d
-    cat > /etc/rancher/$i/config.yaml.d/99-rancheros-labels.yaml << EOF
+    cat > /etc/rancher/$i/config.yaml.d/99-elemental-labels.yaml << EOF
 node-label+:
-- rancheros.cattle.io/image=${IMAGE}
```
mudler, line 11: I hope this isn't relied by anything..

framework/files/usr/sbin/elemental-populate-node-labels
```diff
@@ -6,10 +6,10 @@ set -e
...
-- rancheros.cattle.io/tag=${IMAGE_TAG}
+- elemental.cattle.io/image=${IMAGE}
+- elemental.cattle.io/tag=${IMAGE_TAG}
```
mudler, line 12: @kkaempf do we have the `IMAGE_TAG` in the os-release generated by OBS?

This file is sourcing `rancheros-release` file at the moment.

Itxaka: nope, but we can just move that to be on os-release with a small change so this should be good to go once we change that

mudler:
Only things I see still missing with a grep now are, tests, old scripts used for building and docs. But that would likely be a better fit in a follow-up

Itxaka (approved):

PR #191, opened 2022-07-21
ci: better to use actions-setup-docker to install docker

ldevulder:
Tested here: https://github.com/ldevulder/gha-tests/actions/runs/2710796777


Itxaka (approved):
niiiice

PR #201, opened 2022-07-21
Re add e2e tests

### LINKED ISSUE #182: Re-add E2E tests
See PR #201.

Re-add the E2E tests after changes done in PR #178.

Steps to do:
- [x] update Go modules and Go version to 1.18
- [x] increase k3s version (if needed)
- [x] ~increase Rancher version (if needed)~ (by default latest released version is installed)
- [x] try to fix sporadic timeout issues during installation/bootstrapping tests
- [x] use `elemental-operator` instead of `rancher-operator`
- [x] use latest Elemental configuration/syntax

Re-integration of upgrade tests will be done after.

ldevulder:
Should fix #182 

Verification run: https://github.com/ldevulder/elemental/actions/runs/2774122581

**EDIT of 01/08/2022:** add a 2nd node and multiple reboot check (to ensure that the cluster is still healthy), verification run for this: https://github.com/ldevulder/elemental/actions/runs/2777738676


Makefile
```diff
@@ -106,9 +107,8 @@ endif
 	echo "set initrd elemental-${FINAL_TAG}-initrd" >> build/elemental-${FINAL_TAG}.ipxe
 	echo "set rootfs elemental-${FINAL_TAG}.squashfs" >> build/elemental-${FINAL_TAG}.ipxe
 	echo "set iso    elemental-${FINAL_TAG}.iso" >> build/elemental-${FINAL_TAG}.ipxe  #not used anymore, check if we can boot from iso directly with sanboot?
-	echo "kernel ${url}/${kernel} initrd=${initrd} ip=dhcp rd.cos.disable root=live:${url}/${rootfs} console=tty1 console=ttyS0 ${cmdline}"  >> build/elemental-${FINAL_TAG}.ipxe
-	echo "initrd ${url}${initrd}"  >> build/elemental-${FINAL_TAG}.ipxe
-	echo "boot" >> build/elemental-${FINAL_TAG}.ipxe
+	echo "initrd \$${url}/\$${initrd}"  >> build/elemental-${FINAL_TAG}.ipxe
```
Itxaka, line 110: Any reason for the double $$ and the escape char? This should end up as ${url} in the file no? This is a simple echo no?

I probably mangled it as I have no idea about ipxe files or format

ldevulder: Yes we need to have `${url}` in the ipxe file, it's a variable in the script (like in bash). So current version doesn't work because of this, for example we have `kernel / initr= ip=dhcp [...]` instead of `kernel ${url}/${kernel} initrd=${initrd} ip=dhcp [...]`.

.github/workflows/e2e.yaml
```diff
@@ -1,54 +1,64 @@
-name: e2e
+name: Elemental End-To-End tests with Rancher Manager
 
 on:
-  push:
+  workflow_run:
```
Itxaka, line 4: Why this choice of one workflow running the other instead of being jobs in the same workflow?

tests/e2e/suite_test.go
```diff
@@ -25,13 +25,16 @@ import (
...
+	vmNameRoot         = "node"
 	userName           = "root"
 	userPassword       = "r0s@pwd1"
 	netDefaultFileName = "../assets/net-default.xml"
+	clusterYaml        = "../assets/cluster.yaml"
+	selectorYaml       = "../assets/selector.yaml"
 )
 
 var (
+	k3sVersion  string
```
Itxaka, line 37: Nit: I would change this to kubernetesVersion as this can be any of rke, rkew or keys versions and we might want to matrix them down the line as they are a bit different.


But it's ok now.

Itxaka (approved):

Itxaka:
Awesome job @ldevulder !

.github/workflows/e2e.yaml
ldevulder: For later re-integration of smoke/integration tests, the `build-ci` process will be the same. So I find this syntax better than duplicating the code. Also the E2E tests will not be triggered on each PR but only on merge on `master` (E2E tests are too long - will be more after re-integration of upgrade tests), buit the build could be to have some "quick" validation of the code.

tests/e2e/suite_test.go
ldevulder: Yes that's true, and I already planned to add an RKE2 test, so clearly not a good variable name. What about `k8sVersion`? I found `kubernetesVersion` is a little bit too long ;-)

ldevulder: I renamed the variable to `k8sVersion`. There is still hardcoded references to K3s but it's expected, as  it's for K3s installation. This will be changed as soon as I will add RKE2.

Itxaka (approved):

.github/workflows/e2e.yaml
Itxaka: ah interesting, good choice here!

PR #202, opened 2022-07-22
Add elemental install logs in journal

davidcassany:


davidcassany:
I just tested it and works, even I am not very happy with the solution this a quick and easy fix for the moment. 

IMHO it all turns back to and old discussion we never really solved, which is `yip` logging. We need to enhance that and be capable of distinguishing yip itself logs vs pluging logs  or distinguish certain types of errors, etc. IMHO this would be the proper solution and it would also benefit any kind of cloud-init execution logging.

Itxaka (approved):

PR #203, opened 2022-07-22
Fix custom cloud config while building iso

Itxaka:
Of course we dont have access to that file, we need to copy it if we are
passing a config into the iso/config so ti can be picked up on the image
build, just like on the elemental-iso-build

NOTE: this only affects building local isos from scratch, i.e. not with OBS artifacts so end-users isos should not be affected.


fgiudici (approved):
LGTM!

PR #204, opened 2022-07-22
Improve elemental-operator call

davidcassany:
This commit makes a better use of systemd-cat and it also
makes elemental-operator to be executed on network.after stage.

There is a race condition while calling `elemental install` in
network stage. This is an attempt to avoid it.

### DIFF
--- a/framework/files/system/oem/99_elemental-operator.yaml
+++ b/framework/files/system/oem/99_elemental-operator.yaml
@@ -1,7 +1,7 @@
 name: "Elemental operator bootstrap"
 stages:
-  network:
+  network.after:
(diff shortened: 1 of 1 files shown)


davidcassany:
With this I rarely get rancher/elemental-operator#50 issue, but I still do time to time 😞 in any case I believe this is a little improvement.

Itxaka (approved):

Itxaka:
> With this journalctl -f -t elemental shows elemental-cli logs on real time, I am not sure was working in real time with the pipeline approach.

It was, I tried locally and could see stuff happening in real time!

fgiudici (approved):
LGTM

PR #205, opened 2022-07-22
Bump repo for teal packages

Itxaka:
We cant track a fixed repo on github builds as we would be really far
aways from latest elemental packages

This meant that on github builds, where we install elemental from teal repos instead of OBS, we were reaaaally behind the latest packages


fgiudici (approved):
looks good!

PR #208, opened 2022-07-25
Minor doc fixes

frelon:

### DIFF
--- a/docs/ami.md
+++ b/docs/ami.md
@@ -1,6 +1,6 @@
 # Amazon AMIs
 
-AMIs for RancherOS are published under the owner ID `275947076441` in the `us-west-1` and `us-west-2` regions
+AMIs for Elemental are published under the owner ID `275947076441` in the `us-west-1` and `us-west-2` regions
 currently.
 
 ```bash
--- a/docs/ami.md
+++ b/docs/ami.md
@@ -1,9 +1,0 @@
-# Amazon AMIs
-
-AMIs for Elemental are published under the owner ID `275947076441` in the `us-west-1` and `us-west-2` regions
-currently.
-
-```bash
-aws --region=us-west-1 ec2 describe-images --owners 275947076441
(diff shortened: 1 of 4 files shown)


docs/ami.md
```diff
@@ -1,6 +1,6 @@
 # Amazon AMIs
 
-AMIs for RancherOS are published under the owner ID `275947076441` in the `us-west-1` and `us-west-2` regions
+AMIs for Elemental are published under the owner ID `275947076441` in the `us-west-1` and `us-west-2` regions
```
Itxaka, line 3: I think the Ami stuff can be fully dropped as we are not publishing anything on aws anymore and the owner I'd gas changed for sure

Itxaka (approved):

docs/ami.md
frelon: Nice, will drop the whole thing!

docs/installation.md
```diff
@@ -18,13 +18,13 @@ The installation configuration should be hosted on an HTTP or TFTP server. A sim
 ### Kernel Command Line
 
 Install directives can be set from the kernel command line using a period (.) seperated key structure such as
-`Elemental.install.config_url`.  They kernel command line keys are case-insensitive.
+`elemental.install.config_url`.  They kernel command line keys are case-insensitive.
```
fgiudici, line 21: Is this for cloud-config file passing via kernel cmd line 🤔 ? 
I guess this should be now:
```suggestion
`elemental.install.config-urls`.  They kernel command line keys are case-insensitive.
```
But I think @davidcassany knows better!

fgiudici (approved):
Ugh, we should really start taking care of the docs... thanks @frelon for improving them!

docs/installation.md
frelon: Hmm, that might explain a few errors I've run into! I will try that on a new install and see if it works better!

frelon: Seems like this i deprecated, I removed the whole section instead :+1:

PR #211, opened 2022-07-26
Fix agent dir

Itxaka:
Also enable debug mode by default in the agent so we can see whats going
on


davidcassany (approved):
Looks good to me :+1:

fgiudici (commented):
good catch!

PR #212, opened 2022-07-26
Do not use cdrom provider datasource

davidcassany:
Avoid the use of 'cdrom' provider for datasource plugin. This provider
opens all available block devices to probe them, however an `lsof` call
shows they are not closed afterwards. This might cause issues if within
the same process we pretend to repartition a disk for installation.

This is a workaround for rancher/elemental-operator#50

### DIFF
--- a/framework/files/system/oem/01_elemental-rootfs.yaml
+++ b/framework/files/system/oem/01_elemental-rootfs.yaml
@@ -71,7 +71,7 @@ stages:
     - if: '[ ! -f /oem/userdata ]'
       name: "Pull data from provider"
       datasource:
-        providers: ["cdrom", "gcp", "openstack", "aws", "azure", "hetzner", "packet", "scaleway", "vultr", "digitalocean", "metaldata" ]
+        providers: ["gcp", "openstack", "aws", "azure", "hetzner", "packet", "scaleway", "vultr", "digitalocean", "metaldata" ]
         path: "/oem"
       files:
       - path: /oem/userdata_load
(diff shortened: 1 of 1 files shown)


framework/files/system/oem/01_elemental-rootfs.yaml
```diff
@@ -71,7 +71,7 @@ stages:
     - if: '[ ! -f /oem/userdata ]'
       name: "Pull data from provider"
       datasource:
-        providers: ["cdrom", "gcp", "openstack", "aws", "azure", "hetzner", "packet", "scaleway", "vultr", "digitalocean", "metaldata" ]
+        providers: ["gcp", "openstack", "aws", "azure", "hetzner", "packet", "scaleway", "vultr", "digitalocean", "metaldata" ]
```
mudler, line 74: well, if the edge is elemental's focus, does it make sense to keep data sources at all? The only left here would be just for cloud

mudler (approved):
Looks good here, but I do wonder if keeping this file at all

framework/files/system/oem/01_elemental-rootfs.yaml
Itxaka: kind of agreed. config will come from the embedded file always no? not from any datasources.

Itxaka (approved):

framework/files/system/oem/01_elemental-rootfs.yaml
kkaempf: Then let's drop it !
(Well, at least for now ... we might need to resurrect it at a later point :wink:)

davidcassany: Good, this will also speed up the boot process, now we were waiting to fail fetching data for all cloud providers 😅 

I am completely cleaning this part then.

PR #213, opened 2022-07-27
Enable quotes on /etc/os-release

Itxaka:
Otherwise we get failures when we source the file

### DIFF
--- a/Dockerfile.image
+++ b/Dockerfile.image
@@ -38,22 +38,13 @@ COPY framework/files/ /
 
 ARG IMAGE_TAG=latest
 ARG IMAGE_COMMIT=""
-RUN echo NAME="Elemental" > /etc/os-release
-RUN echo ID="elemental" >> /etc/os-release
-RUN echo ID_LIKE="suse opensuse" >> /etc/os-release
-RUN echo BUG_REPORT_URL="https://github.com/rancher/elemental/issues" >> /etc/os-release
-RUN echo HOME_URL="https://github.com/rancher/elemental" >> /etc/os-release
-RUN echo "VERSION=${IMAGE_TAG}" >> /etc/os-release
-RUN echo "VERSION_ID=$(echo ${IMAGE_TAG} | sed s/^v//)" >> /etc/os-release
-RUN echo "PRETTY_NAME=Elemental ${IMAGE_TAG}" >> /etc/os-release
-RUN echo "COMMIT=${IMAGE_COMMIT}" >> /etc/os-release
-
-# IMPORTANT: Setup os-release used for versioning/upgrade. The
-# values here should reflect the tag of the image being built
 ARG IMAGE_REPO=norepo
-RUN echo "IMAGE_REPO=${IMAGE_REPO}"         >> /etc/os-release
-RUN echo "IMAGE_TAG=${IMAGE_TAG}"           >> /etc/os-release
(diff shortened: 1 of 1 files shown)


Itxaka:
nope, not right yet :/

Itxaka:
Now its fixed!

kkaempf:
Hmm, you're **re**-creating `/etc/os-release` here and thereby overwriting a file owned by the OS.
This is not acceptable for re-distributable images.

Itxaka:
> Hmm, you're re-creating /etc/os-release here and thereby overwriting a file owned by the OS.
This is not acceptable for re-distributable images.

Has been like this since the start, Im just fixing an issue with quotes not being in there `¯\_(ツ)_/¯`

kkaempf:
What's the purpose of `Dockerfile.image` ?
Even if it's just used for CI testing, it might give false test results. /etc/os-release _is_ used by (resp. a plan executed by) e.g. `rancher-system-agent`

Itxaka:
> What's the purpose of Dockerfile.image ?

To build an elemental OCI image? One that can be used by the Makefile.iso to build a final iso? One that can be tested by anyone easily?

> Even if it's just used for CI testing, it might give false test results. /etc/os-release is used by (resp. a plan executed by) e.g. rancher-system-agent

I know, that is why those values are in there, to be used by the system-agent? We do the same on OBS, we override those values on os-release so they can be picked by the populate-node-labels for example?

Im not sure what is your objection to this? This produces a working image and ISO from the source code in this repo. One that can be tested before changes are merged. Should we just drop this? Im not sure how are we expected to not only CI test elemental but develop it? I am using this dockerfiles to test changes to everything, from the final elemental image, to the iso generation, to the cloud-configs and even new versions of the operator, as its easily overridable.

Can you state clearly your objection to this change, which actually fixes an issue in existing code (since the beginning of elemental even)?

Itxaka:
BTW, this does not affect anything on OBS or the official images.

fgiudici (approved):
This change just adds quotes to the values in /etc/os-release, fixing /etc/os-release (quotes are expected there).
So, also if we will decide soon to remove the whole os-release rewrite section in the Dockerfile, applying this change will just fix the format (no change to the vaules) of os-release file. So, I would merge this.
LGTM

kkaempf:
> BTW, this does not affect anything on OBS or the official images.

:eyes:

kkaempf:
> 
> Can you state clearly your objection to this change, which actually fixes an issue in existing code (since the beginning of elemental even)?

/etc/os-release is part of the operating system, used as an identifier on many occasions.

While we can _add_ to it (like adding a TIMESTAMP by the node-image Dockerfile in OBS), we must not rewrite it.

Itxaka:
wait, so its ok if I append a `NAME=whatever` to the file at the end, which will overwrite the original value as it will be loaded later in the file when doing a source `/etc/os-release` ??

I know its an identifier, and we override that as we are constructing a derived work from the original. But Im not sure what does that has anything to do with this PR, and why we must do that, is that on the license agreement of using the dockerfile? 

Would also be ok to NOT copy the file from the original source and just create one ourselves? In that case we would not be modifying anything, but creating the file from scratch?

I just dont see the reasoning, nor what it has to do with this PR, nor why it even matters lol So many workarounds around that that I cant even see how is that an issue.

Itxaka:
@kkaempf removed all uneeded keys and avoided overwriting the file. Now we only add our extra values in there.

kkaempf (approved):
Thank you !

PR #214, opened 2022-07-27
build: remove obsolete elemental.install.* options

ldevulder:
**Note:** iPXE boot script generated in `Makefile` will be fixed in PR #201.


Itxaka (approved):
woops, leftover, good catch!

davidcassany (approved):

PR #216, opened 2022-07-28
Push master images to elemental-ci

Itxaka:
Push the created images from master to the
quay.io/costoolkit/elemental-ci repo to have dev images published so we
can easily build isos from them to test them

### DIFF
--- a/.github/workflows/image.yaml
+++ b/.github/workflows/image.yaml
@@ -4,6 +4,11 @@ on:
   push:
     branches:
       - master
+
+concurrency:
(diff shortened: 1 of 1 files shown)


Itxaka:
This:
 - Makes it easy to go back to a built image in a point in time and inspect it to check if it works or not.
 - Makes it easy to use the same image as we use on CI for troubleshooting
 - Makes it easy to build isos from the latest master branch

frelon (approved):
LGTM!

PR #219, opened 2022-07-28
Improve service unit file

davidcassany:
oneshot service unit type is not active until the executed cmd ends,
however the system agent is a service and polls for plans in an
infinite loop. A simple type is more appropriate, also adding restart
in case something fails contacting to the management cluster.


kkaempf:
Not sure what issue is fixed with that change. 🤔 
Note that elemental-system-agent won't be started once rancher-system-agent is deployed 😉  Maybe this needs to change too ?!

Itxaka (approved):

Issue #25, opened 2022-02-24
rd.neednet=1 is needed on boot in order to install ssh keys


mudler:
In case oem files are copied over into `/oem` (that is, after install) and not loaded by datasources, `rd.neednet=1` is needed to guarantee that the `boot` and `initramfs` stages are executed when network is available, for example to fetch ssh user keys. 

Enabling `rd.neednet=1`, on the other hand, makes configuring network via cloud-init not possible.

This is related to https://github.com/rancher-sandbox/cOS-toolkit/issues/1140 and https://github.com/rancher-sandbox/cOS-toolkit/issues/388

mudler:
We could consider for the time being, for example to _not_ be strict on network when booting on ISO, allowing to customize network settings via cloud-init permanently in the node. Then on booting we can enforce networking with `rd.neednet=1`. At least for the moment being would allow - somehow - to inject network configuration without making complexity around having `rd.neednet=1` disabled. 

Any thoughts @davidcassany ?

davidcassany:
So your suggestion is to make `rd.neednet=0` for the installation ISO and encourage customizations at install time (via hooks or custom cloud-init) to configure network? Sure, this is possible, however I do not see how it solves current issue. I thought the issue was on building generic disks, not ISOs. I thought the entry point (for OpenStack as an example) would be a disk, something that gets dd'ed to the VM actual disk.

Ok another idea comes to my mind, what if we run a two steps setup? I mean we do not include `rd.neednet=1` on recovery and we only add the cdrom datasource for recovery system. Then the recovery system on first boot always runs a reboot to set follow up reboots or the active image parameters. Or some other similar process where we assume firstboot is without network. So on first boot you have the chance to set to add custom cloudinit from cdrom and modify stuff. If no cdrom it just reboots and on the follow up there is `rd.neednet=1`.

mudler:
> So your suggestion is to make `rd.neednet=0` for the installation ISO and encourage customizations at install time (via hooks or custom cloud-init) to configure network? Sure, this is possible, however I do not see how it solves current issue. I thought the issue was on building generic disks, not ISOs. I thought the entry point (for OpenStack as an example) would be a disk, something that gets dd'ed to the VM actual disk.

So it is actually for both - ISO and openstack, correct, the proposal above is just covering ISOs setups scenarios.

[...]

PR #226, opened 2022-07-29
Update bootargs.cfg

Itxaka:
Test disabling rd.neednet

### DIFF
--- a/framework/files/etc/cos/bootargs.cfg
+++ b/framework/files/etc/cos/bootargs.cfg
@@ -2,7 +2,7 @@ set kernel=/boot/vmlinuz
 if [ -n "$recoverylabel" ]; then
     set kernelcmd="console=tty1 console=ttyS0 root=live:LABEL=$recoverylabel rd.live.dir=/ rd.live.squashimg=$img panic=5"
 else
-    set kernelcmd="console=tty1 console=ttyS0 root=LABEL=$label cos-img/filename=$img panic=5 rd.cos.oemlabel=COS_OEM rd.neednet=1"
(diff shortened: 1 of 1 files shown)


kkaempf (approved):
fixes #25 🤞🏻

kkaempf:
How does this relate to https://github.com/rancher/elemental/blob/0a9679fe457322baf3426549ad99811832d9b2eb/tests/smoke/smoke_test.go#L93 ?

Also there's `framework/files/etc/cos/bootargs.cfg` and `framework/cos/etc/cos/bootargs.cfg` and the OBS package build overwrites the latter with the former 🤦🏻

Itxaka:
Ignore the tests, they currently don't run and @ldevulder is working on restoring them.

Itxaka:
Framework/cos is the default files needed from elemental-toolkit and then we override them with the especific elemental ones.

It's on the readme, as it explains that those cos files get auto updated from elemental-toolkit, while the files ones are overlayed on top. Things like the default services from cos (cos-setup-fs for example) are part of the cos framework, while other services like the elemental-system-agent are elemental specific and go into the files folder.

Could probably do with a cleanup and merge of those, not sure what we would lose by just merging them....

ldevulder:
> Ignore the tests, they currently don't run and @ldevulder is working on restoring them.

I'm on the E2E tests, not this smoke on yet, but I can re-add them later for sure. Is these smoke tests were part ofwhat was called Integration Tests?

Itxaka:
> I'm on the E2E tests, not this smoke on yet, but I can re-add them later for sure. Is these smoke tests were part ofwhat was called Integration Tests?

yep, it run the tests under the smoke dir only :)

Issue #176, opened 2022-07-18
Remove cOS from GRUB entries


juadk:
We will have to thing about removing cOS from GRUB entries:
![image](https://user-images.githubusercontent.com/6025636/179512465-94e35364-fbd9-4b50-a190-055b617ab319.png)

Itxaka:
A bit weird because I thought obs build is using the framework files?? So this should show up as "RancherOS" instead of cOS ? @kkaempf are we using the framework files in teh OBS build of elemental? The spec file seems to confirm that...

kkaempf:
`.spec` files don't lie 😉

Itxaka:
mmmhf. `GRUB_MENU_ENTRY` is actually picked from the /etc/os-release :facepalm:

Itxaka:
nope, not usre why we also got that lol, its picked from one of the 4 grub config files... :man_facepalming:

Itxaka:
damn, setting custom names is broken.

This will need https://github.com/rancher/elemental-cli/pull/299 and then a bump to the elemental-cli both on cos-toolkit and obs 

cc @kkaempf

kkaempf:
OBS ✅

PR #227, opened 2022-07-29
Pick elemental from the docker image and set grub name

Itxaka:
For building the image locally/dev/github ci just use a fixed
elemental-cli image to get the elemental-cli binary instead of having to
wait for elemental-toolkit to publsih the packages so we can easily
change versions and test unpublished versions

Also add the GRUB_ENTRY_NAME to os-release so grub looks nicer

Fixes https://github.com/rancher/elemental/issues/176

### DIFF
--- a/Dockerfile.image
+++ b/Dockerfile.image
@@ -49,7 +49,7 @@ RUN echo COMMIT=\"${IMAGE_COMMIT}\" >> /etc/os-release
 RUN echo IMAGE_REPO=\"${IMAGE_REPO}\" >> /etc/os-release
 RUN echo IMAGE_TAG=\"${IMAGE_TAG}\" >> /etc/os-release
 RUN echo IMAGE=\"${IMAGE_REPO}:${IMAGE_TAG}\" >> /etc/os-release
-RUN echo GRUB_ENTRY_NAME=\"Elemental ${IMAGE_TAG}\" >> /etc/os-release
+RUN echo GRUB_ENTRY_NAME=\"Elemental\" >> /etc/os-release
 
 # Rebuild initrd to setup dracut with the boot configurations
 RUN mkinitrd && \
--- a/Dockerfile.image
+++ b/Dockerfile.image
@@ -1,6 +1,7 @@
 # ARGS go first if used on FROM
 ARG OPERATOR_IMAGE=quay.io/costoolkit/elemental-operator:latest
 ARG SYSTEM_AGENT_IMAGE=rancher/system-agent:v0.2.9
+ARG TOOL_IMAGE=quay.io/costoolkit/elemental-cli:v0.0.15-87f0cb4
 # Binaries and files needed from elemental-toolkit repository
 ARG LUET_VERSION=0.32.5
 FROM quay.io/luet/base:$LUET_VERSION AS framework-build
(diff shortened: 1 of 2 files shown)


Itxaka:
cc @kkaempf doesnt affect OBS....

Dockerfile.image
```diff
@@ -1,6 +1,7 @@
 # ARGS go first if used on FROM
 ARG OPERATOR_IMAGE=quay.io/costoolkit/elemental-operator:latest
 ARG SYSTEM_AGENT_IMAGE=rancher/system-agent:v0.2.9
+ARG TOOL_IMAGE=quay.io/costoolkit/elemental-cli:v0.0.15-87f0cb4
```
davidcassany, line 4: @Itxaka  I just have a doubt about how you envision to maintain this, I mean, I assume we simply manually update this value as we need, but is there a some automation to ensure it follows the latest version? or would it make sense to have also publish `quay.io/costoolkit/elemental-cli:latest` so we don't have to worry to pin some specific commit unless we explicitly need to test some specific version.

Just wondering about how this should be handled, any suggestion? In any case like having the option to build locally a built of my choice.

davidcassany (approved):
Looks good :+1:

Dockerfile.image
```diff
@@ -45,6 +49,7 @@ RUN echo COMMIT=\"${IMAGE_COMMIT}\" >> /etc/os-release
 RUN echo IMAGE_REPO=\"${IMAGE_REPO}\" >> /etc/os-release
 RUN echo IMAGE_TAG=\"${IMAGE_TAG}\" >> /etc/os-release
 RUN echo IMAGE=\"${IMAGE_REPO}:${IMAGE_TAG}\" >> /etc/os-release
+RUN echo GRUB_ENTRY_NAME=\"Elemental ${IMAGE_TAG}\" >> /etc/os-release
```
davidcassany, line 52: nit: I  know we used the have it alaways in this way, but I think having the version here is a bad idea. This is going to be used by active and passive grub options. After an update they are likely to be on a different version, however both will be named with the active version (the upgraded one), I find it pretty confusing.

Dockerfile.image
Itxaka: yep, I would like to move this(elemental-cli) and the operator to publish their master images into the `REPO-ci` so we can use those for CI/CD  easily by setting this to that latest tag.

Currently elemental-cli pushes master AND tag versions to the "release" repo which it was good before when it was a moving target but its not needed anymore. Idea would be like operator which pushes master+latest ot the quay.io/costoolkit/elemental-operator-ci and tags+latest to quay.io/costoolkit/elemental-operator repo as to have a clear separation between what is stable and what is dev

Once that is done, this can be moved to track the ci latest tag so we cant test with latest of everything in github and have a good idea if we break something.

Dockerfile.image
Itxaka: that is a good point, although take into consideration that this is only set here for github stuff. Our official releases are all done on OBS now so we probably wont have any releases here anymore as it doesn't make sense. 

I found that having the tag in there helps a lot when testing locally as you can see at a glance what was installed on booting, but I can easily drop, after all same info is available under os-release....maybe it would be even better to show that par under the motd...but that affects also downstream obs builds.

Or maybe I can just leave it out and have the Elemental in there, not really  care one way or the other

Itxaka: just dropped, it. Makes no sense and its closer to OBS

fgiudici (approved):
looks good!

PR #231, opened 2022-08-02
Use the new registration binary to register

Itxaka:
Also introduce for local docker build the REGISTER_IMAGE to be able to
override the wanted register image.

Alos point the operator and register image to the latest master
published images to be on edge so we can catch any issues on changes for
those.


kkaempf (approved):
I was waiting for this 😆

PR #234, opened 2022-08-02
Do not enable elemental-system-agent.service

davidcassany:
We should prevent having enabled services within the image unless we are fully confident we want them to be always enabled and running. Having services enabled just for the bootstrap phase is risky IMHO, as this is written within the immutable filesystem, we should only call the service when needed without enabling it.

### DIFF
--- a/framework/files/system/oem/99_elemental_system_agent.yaml
+++ b/framework/files/system/oem/99_elemental_system_agent.yaml
@@ -1,7 +1,6 @@
 name: "Elemental system agent bootstarp"
 stages:
   network.after:
-    # run only if on live cd and there is a config file
-    - if: '[ ! -f /run/cos/live_mode ] && [ ! -f /etc/systemd/system/rancher-system-agent.service ]'
+    - if: '[ ! -f /etc/systemd/system/rancher-system-agent.service ] && [ ! -f /run/cos/live_mode ] && [! -f /run/cos/recovery_mode ]'
       commands:
         - systemctl start elemental-system-agent.service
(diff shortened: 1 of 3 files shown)


davidcassany:
This is related to rancher/elemental-operator#86

framework/files/system/oem/99_elemental_system_agent.yaml
```diff
@@ -0,0 +1,7 @@
+name: "Elemental system agent bootstarp"
+stages:
+  network.after:
+    # run only if on live cd and there is a config file
```
Itxaka, line 4: coment is wrong :D

framework/files/system/oem/99_elemental_system_agent.yaml
```diff
@@ -0,0 +1,7 @@
...
+    - if: '[ ! -f /run/cos/live_mode ] && [ ! -f /etc/systemd/system/rancher-system-agent.service ]'
```
Itxaka, line 5: what about recovery? shouldnt we just launch this under active partition?

kkaempf:
This needs to be followed-up in https://build.opensuse.org/package/view_file/isv:Rancher:Elemental/systemd-presets-branding-Elemental/50-default-Elemental.preset?expand=1

framework/files/system/oem/99_elemental_system_agent.yaml
davidcassany: good point

Itxaka (approved):
Nice!

davidcassany:
> This needs to be followed-up in https://build.opensuse.org/package/view_file/isv:Rancher:Elemental/systemd-presets-branding-Elemental/50-default-Elemental.preset?expand=1

Here it is https://build.opensuse.org/request/show/992532

PR #238, opened 2022-08-02
E2E: add SetTimeout function for slow runner

ldevulder:
Our self-hosted runner can be slow sometimes (SSD might be needed!), this PR adds a funciton to increase the timeout easily based on a env variable (`TIMEOUT_SCALE`). Set it to 2, should be enough (but will see after testing).

This PR also fixes small issues.

### DIFF
--- a/tests/e2e/suite_test.go
+++ b/tests/e2e/suite_test.go
@@ -54,7 +54,7 @@ func TestE2E(t *testing.T) {
 }
 
 var _ = BeforeSuite(func() {
-	k8sVersion = os.Getenv("INSTALL_K8S_VERSION")
+	k8sVersion = os.Getenv("INSTALL_K3S_VERSION")
 	clusterName = os.Getenv("CLUSTER_NAME")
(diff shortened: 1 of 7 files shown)


tests/e2e/suite_test.go
```diff
@@ -54,7 +54,7 @@ func TestE2E(t *testing.T) {
 }
 
 var _ = BeforeSuite(func() {
-	k8sVersion = os.Getenv("INSTALL_K8S_VERSION")
+	k8sVersion = os.Getenv("INSTALL_K3S_VERSION")
```
ldevulder, line 57: This is normal: variable `INSTALL_K3S_VERSION` is used by K3s installer. I will change this to something more modular when I will add RKE2 test.

Itxaka (approved):

ldevulder:
Just set it to draft, as I saw some syntax error during the validation on my lab: https://github.com/ldevulder/elemental/runs/7633631280?check_suite_focus=true#step:3:955

PR #261, opened 2022-08-08
[WIP] Clean up /etc/resolv.conf to be regenerated

mjura:
In elemental /etc/resolv.conf should be configured as symbolic link to
/etc/resolv.conf -> /run/netconfig/resolv.conf, it is done by wicked or
NetworkManager.

Fixes https://github.com/orgs/rancher/projects/9

### DIFF
--- a/framework/cos/system/oem/05_network.yaml
+++ b/framework/cos/system/oem/05_network.yaml
@@ -7,13 +7,20 @@
 # copy the file with a prefix starting by 90, e.g. /oem/91_custom.yaml
 name: "Default network configuration"
 stages:
-   initramfs:
-     - name: "Setup network"
-       files:
-       - path: /etc/sysconfig/network/ifcfg-eth0
-         content: |
-                  BOOTPROTO='dhcp'
-                  STARTMODE='onboot'
-         permissions: 0600
-         owner: 0
-         group: 0
+  initramfs:
+    - name: "Setup network"
+      files:
+      - path: /etc/sysconfig/network/ifcfg-eth0
+        content: |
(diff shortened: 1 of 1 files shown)


framework/cos/system/oem/05_network.yaml
```diff
@@ -7,13 +7,20 @@
...
+                 BOOTPROTO='dhcp'
+                 STARTMODE='onboot'
+        permissions: 0600
+        owner: 0
+        group: 0
+    - if: '[ ! -f /run/cos/recovery_mode ] && [ ! -L /etc/resolv.conf ]'
+      name: "Remove static resolv.conf"
+      commands:
+      - |
+        if [ -f /etc/resolv.conf ]; then
```
Itxaka, line 24: This should not be necessary, as the -L check already fails if the file does not exists but it doesnt hurt :)

Itxaka (approved):

Itxaka (approved):

Itxaka:
weird, needed to approve it twice for the check to appear :/

framework/cos/system/oem/05_network.yaml
mjura: unfortunately we have to check if `/etc/resolv.conf` exists otherwise in systemd logs we will see that `file does not exist` before upgrade

fgiudici (approved):
lgtm

mjura:
Third version of fix can be found here https://github.com/rancher/elemental/pull/265

mjura:
It is not needed any more, patch https://github.com/rancher/elemental/pull/265 was merged

PR #265, opened 2022-08-09
Copy /etc/resolv.conf from host during upgrade

mjura:
After elemental upgrade we have to also copy /etc/resolv.conf
from /host filesystem, otherwise  it will be taken from
container context.

Fixes https://github.com/orgs/rancher/projects/9


Itxaka (approved):
sucks that it has to be this complicated due to the container context :cry:

PR #270, opened 2022-08-10
Add some basic checks to iso generated by elemental-iso-build

### LINKED ISSUE #195: elemental-iso-build: Test output iso is correct
We could have some smaller tests to check that the generated iso is at least valid

 - Test the script builds with a custom config.yaml
 - Test the config.yaml ends up inside the iso
 - Test the iso is bootable (eltorito)
 - Test the iso contains the kernel and initrd (we have already seen failures on this for different arches)

Should be easy fast tests, nothing major

Itxaka:
Fixes #195 

### DIFF
--- a/.github/workflows/build-iso.yaml
+++ b/.github/workflows/build-iso.yaml
@@ -1,6 +1,7 @@
 name: Test build iso from simple script
 
 on:
+  pull_request:
   push:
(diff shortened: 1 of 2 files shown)


kkaempf (approved):
:+1:

PR #289, opened 2022-08-25
Readd e2e upgrade tests

ldevulder:
This PR should partially fix https://github.com/rancher/elemental/issues/290 by:
- fixing the cleanup, better to not do it in a separate step and add a FULL clean
- re-add some upgrade tests, the manual on as well as the one with `osImage` and with `clusterTargets`, more to come later

Verification run on my lab (with 2 nodes instead of 3, as 3 is too much for my lab): https://github.com/ldevulder/elemental/actions/runs/2927513548


juadk (approved):
LGTM

Issue #297, opened 2022-08-30
Switch (testing) to Teal53


kkaempf:
Use the artifacts from `isv:Rancher:Elemental:Teal53` (instead of `isv:Rancher:Elemental:Teal52`) in testing.

Itxaka:
but testing on github is done using the github artifacts, i.e.:

 - latest elemental-cli from elemental-cli-ci repo
 - latest elemental-operator from elemental-operator-ci repo
 - latest rancher-system-agent from rancher-system-agent repo

Testing on github is only setup for the latest most recent versions published as it should be, that is done in order to catch issues while developing.

Artifacts generated from OBS should be tested in a different way as those can be behind what we have in the edge of development.

Not sure if we should mix obs and github. Cant we have a different way of testing artifacts generated by OBS? Makes no sense to me to mix those.

My thinking is, we develop on github, on main branches, which may not be ready to publish and have breaking changes, and those changes are tested in there -> Once tested, approved and whatnot those changes are propagated to OBS which builds the final image that will be available to customers/end users as "supported" (github stuff is not supported, its dev only). Those artifacts then should be tested before being published.

I mean we can have a daily job that tests the stuff already published on OBS but...how does that makes sense? Its already published adn its not reported anywhere. We cannot have it as part of our dev workflow as any changes done to the code wont be reflected on that job until the source in OBS is updated...by that time it has already been merged to master...

If the stuff is built on OBS, the testing should be done as part of OBS workflows, not github.

Itxaka:
What we can do is change the base testing image which is currently `registry.suse.com/suse/sle-micro-rancher/5.2:latest` to 5.3

kkaempf:
> Not sure if we should mix obs and github

No,I agree we shouldn't.

> What we can do is change the base testing image

That's also ok !

kkaempf:
> If the stuff is built on OBS, the testing should be done as part of OBS workflows, not github.

openQA anyone ? 😉

Itxaka:
Still missing https://github.com/rancher/elemental-toolkit/pull/1594, reopening

Itxaka:
https://github.com/rancher/elemental-toolkit/pull/1594 was merged, closing. We now use packages and elemental with base 5.3

PR #304, opened 2022-08-31
Bump sle micro to 5.3

Itxaka:
Bump the base os image to 5.3

Fixes https://github.com/rancher/elemental/issues/297

### DIFF
--- a/Dockerfile.image
+++ b/Dockerfile.image
@@ -29,7 +29,7 @@ FROM $TOOL_IMAGE as elemental-cli
 
 
 # Base os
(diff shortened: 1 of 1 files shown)


juadk (approved):
LGTM

davidcassany:
@Itxaka we should also change the base image in elemental-toolkit too in https://github.com/rancher/elemental-toolkit/blob/main/values/teal-x86_64.yaml so we get everything build and tested against the new Leap 15.4 and SLE 15 SP4.

Issue #291, opened 2022-08-26
ci: re-add integration tests


ldevulder:
Re-add former integration tests.

ldevulder:
Closed by mistake!

PR #315, opened 2022-09-08
ci/e2e: add TPM emulation test

ldevulder:
Should fix https://github.com/rancher/elemental/issues/291.

These PR needs to be merged before:
- https://github.com/rancher/elemental-operator/pull/169
- https://github.com/rancher/elemental-operator/pull/170
- https://github.com/rancher/elemental-operator/pull/171

Verification run: https://github.com/ldevulder/elemental/runs/8267439630

### DIFF
--- a/tests/e2e/bootstrap_test.go
+++ b/tests/e2e/bootstrap_test.go
@@ -86,6 +86,34 @@ var _ = Describe("E2E - Bootstrapping node", Label("bootstrap"), func() {
 			Expect(numberOfFile).To(BeNumerically(">=", 1))
 		})
 
+		By("Configuring emulated TPM if needed", func() {
+			// Set correct value for TPM emulation
+			value := "false"
+			if emulateTPM == "true" {
+				value = "true"
+			}
+
+			// Patch the yaml file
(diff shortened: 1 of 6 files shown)


tests/e2e/bootstrap_test.go
```diff
@@ -86,6 +86,34 @@ var _ = Describe("E2E - Bootstrapping node", Label("bootstrap"), func() {
...
 
+		By("Configuring emulated TPM if needed", func() {
+			// Set correct value for TPM emulation
+			value := "false"
+			if emulateTPM == "true" {
+				value = "true"
+			}
+
+			// Patch the yaml file
+			err := tools.Sed("emulate-tpm:.*", "emulate-tpm: "+value, emulatedTPMYaml)
```
Itxaka, line 97: we should probably use go templating for this if we expect to expand this further.

Itxaka (approved):

tests/e2e/bootstrap_test.go
ldevulder: You mean something like this: https://blog.gopheracademy.com/advent-2017/using-go-templates?

Anyway yes, this is something we can consider. I have no idea on how it works, but that's a good way to learn.

Issue #263, opened 2022-07-14
build-iso still pulls from quay.io


kkaempf:
`elemental build-iso` should run offline when the base image is available in the local image cache.
Currently it still pulls data from quay.io (grub2 files apparently).

Not sure why this happens. Either `elemental` looks at the wrong places in the image (or not at all :wink:)

Everything needed for ISO building should be provided by [node-image](https://build.opensuse.org/package/show/isv:Rancher:Elemental:Teal52/node-image) and [builder-image](https://build.opensuse.org/package/show/isv:Rancher:Elemental:Teal52/builder-image).

Itxaka:
moved this into elemental as elemental-cli already provides this.

Ok this is what I thougth about.

Because you can set the dirs in the manifest.yaml file we only need the proper artifacts in the TOOL IMAGE, as thats the one that its used to build the iso.
So lets have a package that ships the artifacts that we used to have from the toolkit repos: https://build.opensuse.org/package/show/home:itxaka/elemental-efi
And then on offline build or whatever you want to call it, we can either add on the fly a manifest.yaml file with those dirs so it doesnt need to go to the internet to pick anything, or we can even add flags to the build-iso file to provide those dirs on the command line as flags.


Still, having the build-iso callable from rancher manager is...weird. Like wht is the usecase there, do everything via the UI? Who does that? How would that work? Where is the iso created and stored? Where are the images pulled to? and a hunderd questions more....

As usual with elemental features, this has no use case nor its urgent nor it has a clear workflow or requirements, so IMHO this is low prio, incomplete request. Why not wait until we have a real request for something like this and then discuss if its feasible, how would it work and so on? i.e. plan the feature?

kkaempf:
> Still, having the build-iso callable from rancher manager is...weird. Like what is the usecase there, do everything via the UI?

Yes, exactly this, do everything via the UI.

> Who does that?

Every single Rancher customer. :wink: Managing k8s clusters with a UI is why Rancher Manager exists.

> How would that work? 

You select a `machineRegistration` and have a  "download ISO" button.

> Where is the iso created and stored?

It's created by `elemental-operator` and downloaded to the user's system.

Itxaka:
> You select a machineRegistration and have a "download ISO" button.

Then this should be moved to the operator then :D

IMO this requires a clear workflow in how it would work and the steps required. Also, this would require an UI implementation, Im not sure we know how to do so so it may involve more rancher folks on the UI side.

[...]

PR #316, opened 2022-09-08
Make use or purely OBS artifacts to build ISOs

davidcassany:
This commit includes a manifest.yaml for `elemental build-iso` command so instead of pulling livecd artifacts from the default community repositories it uses equivalent oci builds in OBS.

Related to #263

### DIFF
--- a/manifest.yaml
+++ b/manifest.yaml
@@ -1,0 +1,6 @@
+iso:
+  uefi:
+    - oci:registry.opensuse.org/isv/rancher/elemental/teal53/15.4/rancher/elemental-live-grub2-efi-image/5.3
+  image:
+    - oci:registry.opensuse.org/isv/rancher/elemental/teal53/15.4/rancher/elemental-live-grub2/5.3
+    - oci:registry.opensuse.org/isv/rancher/elemental/teal53/15.4/rancher/elemental-live-grub2-efi-image/5.3
--- a/Dockerfile.iso
+++ b/Dockerfile.iso
@@ -7,7 +7,6 @@ FROM $TOOL_IMAGE as tools
 FROM tools AS default
 WORKDIR /iso
 COPY --from=os / rootfs
-COPY iso/grub.cfg overlay/boot/grub2/grub.cfg
 ARG CLOUD_CONFIG_FILE=iso/config
 COPY $CLOUD_CONFIG_FILE overlay/livecd-cloud-config.yaml
 ARG ELEMENTAL_VERSION=""
--- a/elemental-iso-build
+++ b/elemental-iso-build
@@ -12,7 +12,6 @@ build() {
     cp "$CLOUD_CONFIG_FILE" "$TEMPDIR"/iso/config
   fi
   curl https://raw.githubusercontent.com/rancher/elemental/main/Dockerfile.iso -fsSL -o "$TEMPDIR"/Dockerfile
-  curl https://raw.githubusercontent.com/rancher/elemental/main/iso/grub.cfg -fsSL -o "$TEMPDIR"/iso/grub.cfg
   # Did we copy the custom cloud config? Then no need to download the default one
   if [[ -z "$CLOUD_CONFIG_FILE" ]]; then
     echo "No custom registration set, using default"
@@ -25,7 +24,7 @@ build() {
 
   rm -Rf "$TEMPDIR"
   ISONAME="elemental-$(date "+%FT%TZ")"
-  docker run --rm -v "$PWD":/mnt elemental/iso:latest build-iso -o /mnt --squash-no-compression -n "$ISONAME" --overlay-iso overlay dir:rootfs
+  docker run --rm -v "$PWD":/mnt elemental/iso:latest --config-dir /mnt build-iso -o /mnt --squash-no-compression -n "$ISONAME" --overlay-iso overlay dir:rootfs
   echo "Iso generated at $ISONAME.iso"
 }
 


Itxaka (commented):
will probably need to remove the line 15 from Dockerfile.iso as that was meant to be set to the elemental version of the iso being built and shows up in the grub menu. Was an easy way to check if the iso was built correctly based on the git version but unused everywhere else so it can be dropped.

Also, just remove the grub.cfg if its not used? it kind of sucks because I did use that to test grub stuff for the iso (like the above version stuff) but its ok, I guess I will need to rework the local stuff and the github stuff.

manifest.yaml
```diff
@@ -0,0 +1,6 @@
+iso:
+  uefi:
+    - oci:registry.opensuse.org/isv/rancher/elemental/teal53/15.4/rancher/elemental-live-grub2-efi-image/5.3
```
Itxaka, line 3: this dont exist yet do they?

manifest.yaml
```diff
@@ -0,0 +1,6 @@
...
+  image:
+    - oci:registry.opensuse.org/isv/rancher/elemental/teal53/15.4/rancher/elemental-live-grub2/5.3
```
Itxaka, line 5: cant we bundle all of the artifacts into just one package?

davidcassany:
I managed to build and use containers inside OSB to substitute the former `live/grub2` and `live/grub2-efi-image` packages from elemental-toolkit.

It turned out that we can actually create containers from scratch in OBS using Dockerfiles, I was under the impression it would not be functional, but it is instead. See (live/grub2)[https://build.opensuse.org/package/show/isv:Rancher:Elemental:Teal53/live-grub2] OBS build, it has damn simple Dockerfile, way simpler than what I initially expected.

I marked it as a draft because this approach is not yet complete, there few things to clarify:

* How to relate the OBS build packages with elemental-toolkit. It is confusing and frustrating having to keep duplicates with completely different sources to build the same. Is there a way to keep them in sync at least?
* How include in the same docker file builds for multiple archs. Currently the packages I created are only for x86_64... I can't see a way apply some sort of templating mechanism based on arch in Dockerfiles... probably there is, but I am not aware.

manifest.yaml
davidcassany: yes they do https://build.opensuse.org/package/show/isv:Rancher:Elemental:Teal53/live-grub2

manifest.yaml
davidcassany: probably, but note we still need to build both in OBS as on is for the EFI image and the efi image contents are also required into the ISO filesystem.

This is just a draft, I am far from happy with this approach, but at least the OBS containers don't make my eyes bleed with strange and weird scripts.

manifest.yaml
Itxaka: ah just saw the packages! arm64 is failing thougth. Spec file is gonna be a pita, Im happy its not me doing them :D

davidcassany: Another options could be to build the live/grub2 as containers as part of the `elemental-iso-build` script and somehow render the Dockerfile according to the arch... 🤔 This could work but turns `elemental-iso-build` script into a little monster (probably not that little...)

ldevulder:
@davidcassany as you are modifying `Dockerfile.iso` in this PR, maybe it could be good to update the images' version as well:
```
diff --git a/Dockerfile.iso b/Dockerfile.iso
index 26b934f..49d6460 100644
--- a/Dockerfile.iso
+++ b/Dockerfile.iso
@@ -1,5 +1,5 @@
-ARG OS_IMAGE=registry.opensuse.org/isv/rancher/elemental/teal52/15.3/rancher/elemental-node-image/5.2:latest
-ARG TOOL_IMAGE=registry.opensuse.org/isv/rancher/elemental/teal52/15.3/rancher/elemental-builder-image
+ARG OS_IMAGE=registry.opensuse.org/isv/rancher/elemental/teal53/15.4/rancher/elemental-node-image/5.3:latest
+ARG TOOL_IMAGE=registry.opensuse.org/isv/rancher/elemental/teal53/15.4/rancher/elemental-builder-image:latest
 
 FROM $OS_IMAGE AS os
 FROM $TOOL_IMAGE as tools
```

davidcassany:
> @davidcassany as you are modifying `Dockerfile.iso` in this PR, maybe it could be good to update the images' version as well:

Indeed this should be updated, however note this PR is just a draft. I am not satisfied with it... This approach might be _functional_ for now, but it looks like we are just abusing of the cracks in OBS Dockerfile support. I doubt it is a good practice using multistage builds in OBS, to me it feels like the Dockerfile parser simply doesn't have any notion of them and hence ignoring them.

kkaempf:
We need to use different image `BuildTag`s to keep x86_64 and aarch64 apart.

kkaempf:
I'm not sure why we need a `live-grub2` image in OBS. Both grub2 packages are already part of `node-image` 🤔 
(and `syslinux` can be easily added)

kkaempf:
Same for `live-grub2-efi-image`

davidcassany:
> I'm not sure why we need a `live-grub2` image in OBS. Both grub2 packages are already part of `node-image` thinking (and `syslinux` can be easily added)

So the reason being of this packages is the logic found in https://github.com/rancher/elemental-toolkit/blob/84c131256ba02de5934b0e337dfb638ec6ef222b/packages/livecd/grub2/build.yaml#L5-L44 This basically creates the grub2 image to be used for ISOs and arranges grub2 binaries and files beyond grub.cfg file.

This could be included as part of the `elemental build-iso` command and relay on finding the artifacts within the OS image, that's currenlty the alternative I am exploring. The counter part is that `elemental build-iso` command is then coupled with some specific installation of SUSE packages (grub2, syslinux and friends). The problem is that each distro cooks grub2 in its own name and equivalent binaries are installed in different locations with different names.

To work around that we simply created a luet package including the bootloader files we need in ISO's root and the ones we need in EFI's image root, then the  `elemental build-iso` command is dumb and simply includes its content. This is allowing us to create ISOs in elemental-toolkit for green, blue and orange flavors all by using the openSUSEs bootloader artifacts as they were simply decoupled form the main OS. As a bonus, this setup also allows as to build cross arch ISOs, since the `elemental build-iso` does not execute any binary, it simply copies data to the right places and burns the ISO.

I agree this current PR is unsatisfying, because it is simply playing around with the contents of some packages without adding anything special and having to distribute images like the ones I created in this PR over the SUSE's registry feels just wrong.

kkaempf:
Understood and agreed on the multi-distro goal.

- can we drop grub related package from node-image now ?
- we need to consider ARM on OBS (not immediately, though)

davidcassany:
> Understood and agreed on the multi-distro goal.
> 
>     * can we drop grub related package from node-image now ?

No, we still need them during the installation to disk, grub2 install on disk is another battle.
 
>     * we need to consider ARM on OBS (not immediately, though)

Sure, however I more towards not building these bootloader artifacts in OBS and arrange them at ISO build time now. We should discuss it during the planning.

PR #321, opened 2022-09-12
Enable networkmanager

Itxaka:

### DIFF
--- a/framework/files/system/oem/99_networking.yaml
+++ b/framework/files/system/oem/99_networking.yaml
@@ -1,0 +1,7 @@
+name: "Enable NetworkManager"
+stages:
+  initramfs:
+    - name: "Enable NetworkManager"
+      systemctl:
+        enable:
+          - NetworkManager


Itxaka:
This fixes it but unfortunately this kind of collides with OBS cc @kkaempf 

Like this is the way we should be defining our services, but on OBS we use the system presets for this...I dont think there is a way we can set the same enable toggle for both services....

And without this, we basically lose our testing ability for development.

We could patch this in only on the github run but then...development wise is a pita as you have to remember to locally patch this. And then it starts to look like OBS with different .patch files for stuff 


:/

no idea what to do.

davidcassany:
Well, IMHO the solution would be to enable it as part of the Dockerfile so this service is actually enabled in the image, not in the ephemeral data created at boot. Then it would be really equivalent to the OBS build.

Another option is to make this as part of the base image. Meaning the package `systemd-presets-branding-Elemental` is included within the sle-micro-for-rancher image.

ldevulder (approved):
As you said, maybe better to find another solution to deal with OBS stuff, but for now looks good as a workaround to not block CI.

kkaempf:
> Another option is to make this as part of the base image. Meaning the package `systemd-presets-branding-Elemental` is included within the sle-micro-for-rancher image.

I avoided that on purpose to give us more flexibility (and quicker turnarounds :wink:)

kkaempf:
> This fixes it but unfortunately this kind of collides with OBS cc @kkaempf

How does it collide ?
It's just a symlink. Creating it with `ln -sf` should not fail if it already exists. 🤔

ldevulder:
FYI I successfully tested this PR on my local runner: https://github.com/ldevulder/elemental/actions/runs/3039498888

Itxaka:
> > This fixes it but unfortunately this kind of collides with OBS cc @kkaempf
> 
> How does it collide ?
> It's just a symlink. Creating it with `ln -sf` should not fail if it already exists. 🤔 

Mostly that there is two points where this is enabled, so if in the future we drop networkmanager we also need to disable in here. Just a pita to have two places to set this that can lead to slips

Itxaka:
Let's merge this, the. It can be discussed and or planned tomorrow in the meeting

PR #327, opened 2022-09-13
boot from shim on efi systems

Itxaka:
efipackage is teh extracted data from the package generated on the toolkit for testing.

Package source is https://github.com/rancher/elemental-toolkit/pull/1604

This creates an iso that is bootable on secureboot systems (with an `ovmf-x86_64-suse-4m-code.bin` firmware)
NOT tested on any other firmware (it may boot because the suse firmware already contains the suse keys, which is what the shim is using IIRC)
NOT working after install. The installer still doesnt add the boot entry to efi so it needs to be manually added.


Itxaka:
>  NOT working after install. The installer still doesnt add the boot entry to efi so it needs to be manually added.

well, wtf, this actually is not true, the installed system boots afterwards with no issue?? at least on qemu...

Itxaka:
> well, wtf, this actually is not true, the installed system boots afterwards with no issue?? at least on qemu...

well yeah I disabled secureboot :D

Itxaka:
ok, so with some changes its possible to use the shim to boot from our grub but...[0]

 - We need to sign the grub.efi generated
 - We need to sign the grub config
 - We need to sign the kernel+initrd
 - We need to enroll the signature with mokmanager

Seems a bit out of our reach currently...

I would just rework a bit the grub stuff to set the proper shim and the bootloader entry and then check with sec team to see how they do this stuff. Im guessing because they dont generate the grub.efi on the fly, they can be properly signed?

There is also another approach and that is to self-sign. This is something that nvidia modules do when you install them, they generate a cert on the fly, sign the modules and store the cert so its enrolled on the next reboot. Kind of crappy and defeats the purpouse of secure boot.

In any case doesnt seems like something that we can tackle easily without  some feedback from teams that know about this.

Itxaka:
checking even further, seems like only our grub.efi which is generated during install is not signed. So the possibility of generating a cert, signing the grub.efi and enrolling the key during install is more probable I guess?

Itxaka:
This has served its function. This show that with a proper updated package of lib/grub we can secureboot the liveCD

Issue #298, opened 2022-08-30
e2e: add upgrade test with ManagedOSVersionChannel


ldevulder:
Next step after https://github.com/rancher/elemental/issues/290

- [x] re-add upgrade with managedOSVersionName and nodeSelector
- [x] fix doc

ldevulder:
Docs should be updated also!

PR #329, opened 2022-09-15
ci/e2e: re-add upgrade with managedOSVersionName

ldevulder:
This test will also use nodeSelector to restrict the upgrade to a specified node.

Should fix https://github.com/rancher/elemental/issues/298.

Verification run on my lab: https://github.com/ldevulder/elemental/actions/runs/3084596304


juadk (approved):
LGTM

PR #337, opened 2022-09-20
Fix ISO build

davidcassany:
Fixes rancher/elemental-cli#320


Dockerfile.iso
```diff
@@ -1,4 +1,4 @@
-ARG OS_IMAGE=registry.opensuse.org/isv/rancher/elemental/teal52/15.3/rancher/elemental-node-image/5.2:latest
+ARG OS_IMAGE=registry.opensuse.org/isv/rancher/elemental/teal53/15.4/rancher/elemental-node-image/5.3:latest
```
davidcassany, line 1: Looks like we have been using the wrong image for a while 🤦🏽

elemental-iso-build
```diff
@@ -11,7 +13,7 @@ build() {
     echo "Using custom registration from $CLOUD_CONFIG_FILE"
     cp "$CLOUD_CONFIG_FILE" "$TEMPDIR"/iso/config
   fi
-  curl https://raw.githubusercontent.com/rancher/elemental/main/Dockerfile.iso -fsSL -o "$TEMPDIR"/Dockerfile
+  cp "${BASEDIR}/Dockerfile.iso" "${TEMPDIR}/Dockerfile"
   curl https://raw.githubusercontent.com/rancher/elemental/main/iso/grub.cfg -fsSL -o "$TEMPDIR"/iso/grub.cfg
```
davidcassany, line 17: This should not be needed as the elemental-cli already includes the grub.cfg and no download should be needed. However for now I kept it since it includes some weird logic to set a version in a grubenv, I have no idea why we need this, I'd drop it. However I prefer doing it in a separate PR, this is just to fix the broken build. Cosmetic details on grub setup specific to this repository will come later.

elemental-iso-build
```diff
@@ -25,7 +27,7 @@ build() {
 
   rm -Rf "$TEMPDIR"
   ISONAME="elemental-$(date "+%FT%TZ")"
-  docker run --rm -v "$PWD":/mnt elemental/iso:latest build-iso -o /mnt --squash-no-compression -n "$ISONAME" --overlay-iso overlay dir:rootfs
+  docker run --privileged --rm -v "$PWD":/mnt elemental/iso:latest build-iso -o /mnt --squash-no-compression -n "$ISONAME" --bootloader-in-rootfs --overlay-iso /iso/overlay dir:rootfs
```
davidcassany, line 30: This --privileged is needed to build legacy firmware ISOs.... as we need to call `grub2-mkimage` in a chroot for that. I am wondering to completely remove BIOS support when using the `--bootloader-in-rootfs` flag, since the previous strategy of pulling artifacts from somewhere else is still possible it will not be a feature loss in any case.

.github/workflows/build-iso.yaml
```diff
@@ -16,12 +16,14 @@ jobs:
...
+      - name: Checkout code
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
       - name: Build iso from script
         run: |
           # Just a simple run to test the build script, no need to upload the generated iso!
           echo "TEST" > config.yaml
-          curl -fsSL https://raw.githubusercontent.com/rancher/elemental/main/elemental-iso-build -o elemental-iso-build
-          chmod +x ./elemental-iso-build
```
davidcassany, line 24: @Itxaka this logic we done from some specific purpose I can't see? This prevents testing script changes on PRs since it always pulls the script from `main`, it does not execute the script form the PR.

I am wondering if this was done for some specific reason I can't see or it is just a leftover.

davidcassany:
We have the failure because in our Dockerfile.image the base image does not include `syslinux`... it will be a mess to include it there, I prefer waiting the base image actually includes it. Gonna disable the offending tests part... for now 😞

kkaempf (changes requested):
ISO building works now, but it fails to boot with
```
'/boot/kernel.xz' not found
```

Dockerfile.iso
ldevulder: Yes, that was my comment on a previous PR where you were modifying this file ;-)

elemental-iso-build
Itxaka: The grub.cfg file was in there since the start and was copied into the image always.

The version stuff can be dropped, this was for dev purposes and should have no bearing in the final code. In fact we do the same on OBS with an empty version IIRC.

davidcassany:
> ISO building works now, but it fails to boot with
> 
> ```
> '/boot/kernel.xz' not found
> ```

I'd say this is problem of the docker builder cache which does not detect a change in the copied files (the dockerfile remains the same) and does not rebuild the image. I just did `docker builder prune --all --force` and the iso builder image got recreated again, then it included the updated grub.cfg.

elemental-iso-build
Itxaka: that is kind of bad, should not need --privileged to build an iso :/

Is there any way we can circumvent that?

.github/workflows/build-iso.yaml
Itxaka: That job is not there to prevent broken scripts getting in, it was just added as a extra to test the validity of the script on main. If you check you will see that it doesn't even checkout the source code, so its like an external user getting the script. 

Please add a different script to check the validity of the modified script instead as it has a different scope.

elemental-iso-build
davidcassany: Yes, I am on it. Building for EFI or BIOS firmware images, but not for both... Gonna fix it a follow up PR though. Only BIOS requires privileges.

Itxaka: we should probably look into dropping bios shortly blargh

kkaempf:
I'm actually not pulling the Dockerfile.iso from GH but have my local copy (which is identical to the `main branch` one, except that it's based on Teal53)

Itxaka (changes requested):
- missing `--privileged` in makefile for local building
 - missing `--bootloader-in-rootfs`in makefile for local building

Itxaka:
> This is just wrong, sorry. This is a chicken and egg problem... the grub config can't be updated in a PR since the script downloads it from main branch... 🤦🏽 Ok, fixing that now...

This is not really a problem. This is a script meant to be used by external users which dont want to clone the repo or anything. That script should only be considered for releases, as it shouldn't be used in any other place. Basically its an extra for non-technical users.

For testing we have the make iso(using local artifacts) and make proper_iso (using OBS artifacts) which again test different things, make iso is mostly for dev envs and local testing, while proper_iso is for testing with remote OBS produced artifacts.

Testing with `make build && make iso` will fail as the `make build` does not install syslinux in the rootfs. `make proper_iso` should work as long as the OBS artifacts have the syslinux packages in them.

Maybe there is a disconnection here as the changes to grub here are the ones that are propagated to the OBS artifacts, and those are the ones then used by the `make proper_iso` but until we merge those grub changes and OBS produces new artifacts, we cannot use `proper_iso`.

The solution here is to add the syslinux packages into our dev image in github (Dockerfile.image) and test the grub changes with those. 

Would be even better if those packages were added on sle-micro-rancher as we wouldn't need to hack our way around.

Thankfully we dont usually need to add packages  because its a trip lol

elemental-iso-build
```diff
@@ -3,18 +3,18 @@
...
   mkdir "$TEMPDIR"/iso
   # passing a config file? Copy it to the target that will get copied inside the build
   if [[ -n "$CLOUD_CONFIG_FILE" ]]; then
     echo "Using custom registration from $CLOUD_CONFIG_FILE"
-    cp "$CLOUD_CONFIG_FILE" "$TEMPDIR"/iso/config
+    cp "$CLOUD_CONFIG_FILE" "${TEMPDIR}/iso/config"
   fi
   cp "${BASEDIR}/Dockerfile.iso" "${TEMPDIR}/Dockerfile"
-  curl https://raw.githubusercontent.com/rancher/elemental/main/iso/grub.cfg -fsSL -o "$TEMPDIR"/iso/grub.cfg
+  cp "${BASEDIR}/iso/manifest.yaml" "${TEMPDIR}/iso/manifest.yaml"
```
Itxaka, line 17: this breaks the whole point of this script, which is for external users to run the script with a config file and get an iso with no cloning of the repo.

elemental-iso-build
```diff
@@ -27,7 +27,7 @@ build() {
 
   rm -Rf "$TEMPDIR"
   ISONAME="elemental-$(date "+%FT%TZ")"
-  docker run --privileged --rm -v "$PWD":/mnt elemental/iso:latest build-iso -o /mnt --squash-no-compression -n "$ISONAME" --bootloader-in-rootfs --overlay-iso /iso/overlay dir:rootfs
+  docker run --privileged --rm -v "$PWD":/mnt elemental/iso:latest --config-dir . build-iso -o /mnt -n "$ISONAME" --overlay-iso /iso/overlay dir:rootfs
```
Itxaka, line 30: please dont drop the `--squash-no-compression` as it saves a LOT of time in building isos, especially on github :)

davidcassany:
> This is not really a problem. This is a script meant to be used by external users which dont want to clone the repo or anything. That script should only be considered for releases, as it shouldn't be used in any other place. Basically its an extra for non-technical users.

I am fine with this approach, but then we should remove the script testing form the PR given that it does not make use of the source form the PR itself. Gonna re-add the files download form _main_ branch then.
 
> For testing we have the make iso(using local artifacts) and make proper_iso (using OBS artifacts) which again test different things, make iso is mostly for dev envs and local testing, while proper_iso is for testing with remote OBS produced artifacts.

Ok I see.
 
> Testing with `make build && make iso` will fail as the `make build` does not install syslinux in the rootfs. `make proper_iso` should work as long as the OBS artifacts have the syslinux packages in them.

Yes definitely, this will fail until syslinux is in the rootfs or until we are capable to build pure EFI images.
 
> Maybe there is a disconnection here as the changes to grub here are the ones that are propagated to the OBS artifacts, and those are the ones then used by the `make proper_iso` but until we merge those grub changes and OBS produces new artifacts, we cannot use `proper_iso`.

Not really, `proper_iso` should be already functional

> 
> The solution here is to add the syslinux packages into our dev image in github (Dockerfile.image) and test the grub changes with those.

I will not do that, it will be really messy. There are not sle micro repositories available as far as I know, so the image comes with no repositories enabled. I don't want to start adding third party binaries here or define opensuse repos on a sle-micro image.

> 
> Would be even better if those packages were added on sle-micro-rancher as we wouldn't need to hack our way around.

definitely

kkaempf:
`syslinux` is part of the OBS node-image/5.3 and has been requested for SLE-Micro-for-Rancher

Itxaka:
> I am fine with this approach, but then we should remove the script testing form the PR given that it does not make use of the source form the PR itself. Gonna re-add the files download form main branch then.

Yes. That was actually a mvp script. It needs to be fully reworked as we may want to include it on the releases (so people can download just that) and that would need to have a fixed tag that we would need to generate. So yeah, it was there becasue it was requested but the bare minimum was done with it. Didnt even noticed that it runs on PR lol, probably a leftover of my testing while adding it :D

> I will not do that, it will be really messy. There are not sle micro repositories available as far as I know, so the image comes with no repositories enabled. I don't want to start adding third party binaries here or define opensuse repos on a sle-micro image.

Well we kind of already do that, dont we? For our local testing, as we are building the images ourselves, we rely on luet to provide some needed packages on top of the OS (we could probably drop all of them LMAO). Talking about luet packages here, not usual RPM packages.

If we dont do that, dev is stopped until the package trickles down into SLE-Micro-for-Rancher as github uses that as base. No e2e tests, no local dev of anything that needs an iso, as our local changes do not get into the iso for testing. No way of testing anything or developing as we wont be able to create an iso with the current setup.

We can use the remote iso but that doesnt include any local changes done here. :shrug:

elemental-iso-build
```diff
@@ -25,7 +27,7 @@ build() {
 
   rm -Rf "$TEMPDIR"
   ISONAME="elemental-$(date "+%FT%TZ")"
-  docker run --rm -v "$PWD":/mnt elemental/iso:latest build-iso -o /mnt --squash-no-compression -n "$ISONAME" --overlay-iso overlay dir:rootfs
+  docker run --privileged --rm -v "$PWD":/mnt elemental/iso:latest --config-dir . build-iso -o /mnt -n "$ISONAME" --overlay-iso /iso/overlay dir:rootfs
```
Itxaka, line 30: I wonder why it needs the full path to the overlay now...is it due to the chroot?

Itxaka:
Im still baffled on where the grub.cfg file for the live iso comes from on this patch? I dont see where it can come from, there is no grub.cfg in the framework and the one specifically to build the iso is gone. And I dont see it being generated as part of the build-iso or anything?

what Im I missing?

Itxaka:
> Im still baffled on where the grub.cfg file for the live iso comes from on this patch? I dont see where it can come from, there is no grub.cfg in the framework and the one specifically to build the iso is gone. And I dont see it being generated as part of the build-iso or anything?

Ah finally found it, was looking in the wrong place :D

elemental-iso-build
davidcassany: Yes I am also so puzzled because of this... 😞

elemental-iso-build
davidcassany: This is part of the manifest.yaml now. I placed there the non-dynamic flags.

davidcassany:
@kkaempf @Itxaka ok the current patch should let you run `make proper_iso` without issues. `make build && make iso` also works since I finally opted to include syslinux from Leap in Dockerfile.image (god forgive... ). At least this way the ISO creation should be unblocked.

`elemental-iso-build` script fails because it will attempt to download files from main branch that are not yet in main branch.

.github/workflows/build-iso.yaml
davidcassany: Finally kept as it was

elemental-iso-build
davidcassany: reverted

ldevulder (approved):
ISO compilation is fixed, I think that's enough for that PR :-) Let's fix the other stuffs in new PR.

Itxaka (approved):

Dockerfile.image
```diff
@@ -44,6 +44,13 @@ COPY --from=system-agent /usr/bin/rancher-system-agent /usr/sbin/elemental-syste
 # Copy elemental
 COPY --from=elemental-cli /usr/bin/elemental /usr/bin/elemental
 
+# DELETEME once syslinux is added in to the base image or no longer needed
+RUN zypper ar --refresh http://download.opensuse.org/distribution/leap/15.4/repo/oss/ Main && \
+    rpm --import http://download.opensuse.org/distribution/leap/15.3/repo/oss/gpg-pubkey-3dbdc284-53674dd4.asc && \
```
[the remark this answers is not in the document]
ldevulder: For a workaround what is the benefit? Anyway if it's a quick change that's fine.

I'm saying this because this issue is blocking the CI and also manual testing with ISO, so if we could have a fix it could be great :-)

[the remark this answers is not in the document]
Itxaka: mainly just wondering why he choose this over adding our toolkit package, knowing that it was used for the isos before(I think?)

Maybe our package is broken or not valid for this workaround. 

Note that this is just a question, no blocking the PR, I actually approved it ;)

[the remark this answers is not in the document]
Itxaka: oh shit, we have the resolve conversations before merging enabled so that is why this wasnt merged automatically. Sorry about that, didnt thougth we had that enabled anywhere as its a pita. Resolving this comment so it gets merged

Itxaka:
we also had a pending changes requested. I thougth those got cleared on new commits :/

juadk:
I have an issue with the new version of the script:
```
vagrant@k3s-opensusetw:~> ./elemental-iso-build my-registration.yaml
Using custom registration from my-registration.yaml
curl: (3) URL using bad/illegal format or missing URL
```

kkaempf:
Confirmed fixed. Running `elemental-iso-build` produces a working ISO again. 
Thanks !

PR #341, opened 2022-09-21
Fix script and enhance it

Itxaka:
- Drop BASENAME as its not used
 - Fix duplicated arg on curl manifest
 - Add TAG var to be able to override where the artifacts come from


Itxaka:
This should allow us to replace the TAG also in releases and provide a script with the tag embedded with each release.

ldevulder (approved):

PR #342, opened 2022-09-21
ci: Add test for elemental-iso-build on PR

Itxaka:
Now that we can modify where the artifacts are obtained from in the
script, we can tests under a PR context the validity of the script.

Also moves the script under the .github dir as it should not be on a
public facing place. Releases will be used to generate a script with the
proper tag embedded so it uses the proper tag to obtain the artifacts


Itxaka:
Also added a comment and changed the name of the main script to make it more clear that the test is done from a external point of view, only accessing the script directly, no other interaction with the repo needed.

ldevulder (approved):

PR #348, opened 2022-09-22
Run build-iso in non privileged containers

davidcassany:

### DIFF
--- a/.github/elemental-iso-build
+++ b/.github/elemental-iso-build
@@ -27,7 +27,7 @@ build() {
 
   rm -Rf "$TEMPDIR"
   ISONAME="elemental-${TAG}-$(date "+%FT%TZ")"
-  docker run --privileged --rm -v "$PWD":/mnt elemental/iso:latest --config-dir . build-iso -o /mnt -n "$ISONAME" --overlay-iso /iso/overlay dir:rootfs
+  docker run --rm -v "$PWD":/mnt elemental/iso:latest --config-dir . build-iso -o /mnt -n "$ISONAME" --overlay-iso overlay dir:rootfs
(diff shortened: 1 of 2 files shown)


davidcassany:
@juadk @ldevulder heads up that from now on ISO images will be built only for EFI firmware by default. Do we need bios firmware for some of the current tests? If so is it feasible to move it to EFI or better keep building BIOS ISOs for testing for now? 

It is possible to build ISOs with BIOS support, but requires an specific flag to `elemental build-iso` and if running in a container this needs to be privileged.

ldevulder:
The E2E tests are using BIOS in QEMU for now. I could try to have a look to move to EFI quickly. I remember that I had issues last time I tried, but it was some months ago.

davidcassany:
> The E2E tests are using BIOS in QEMU for now. I could try to have a look to move to EFI quickly. I remember that I had issues last time I tried, but it was some months ago.

Just added an extra commit so `make build` builds the ISO for BIOS firmware. This should make our CI happy again as far as I understand.

davidcassany:
@ldevulder Ok so for now the `make iso` still builds a BIOS ISO.  `make proper_iso` does it for EFI only. I believe this is good enough for now so we do not break the CI and force the EFI adoption. I guess we should first verify our CI works with EFI and then do the move.

ldevulder:
@davidcassany in fact I made a mistake: current E2E tests are not directly booting the ISO, iPXE boot is used instead. But in BIOS, not EFI. Don't know if this makes a difference for you. But as long as the ipxe files are here I think it's enough for the E2E tests. In that case BIOS boot support from ISO could be removed.

Issue #312, opened 2022-09-06
Create an Elemental issue template


davidcassany:
Create a template issue for Elemental repository

* Should try to avoid elemental-toolkit related issues here

PR #356, opened 2022-09-23
Add Elemental issue templeate

mjura:
Fixes https://github.com/rancher/elemental/issues/312

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,23 @@
+---
+name: Bug report
+about: Tell us about the problem you are experiencing
+
+---
+
+**What steps did you take and what happened:**
+[A clear and concise description of what the bug is.]
+
+
+**What did you expect to happen:**
+
+
+**Anything else you would like to add:**
+[Miscellaneous information that will assist in solving the issue.]
+
+
+**Environment:**
+
+- Elemental release version (use `cat /etc/os-release`):
+- Rancher version:
+- Kubernetes version (use `kubectl version`):
+- Cloud provider or hardware configuration:
--- a/.github/ISSUE_TEMPLATE/feature_request.md
+++ b/.github/ISSUE_TEMPLATE/feature_request.md
@@ -1,0 +1,20 @@
+---
+name: Feature enhancement request
+about: Suggest an idea for this project
+
+---
+
+**Describe the solution you'd like:**
+[A clear and concise description of what you want to happen.]
+
+
+**Anything else you would like to add:**
+[Miscellaneous information that will assist in solving the issue.]
+
+
+**Environment:**
+
+- Elemental release version (use `cat /etc/os-release`):
+- Rancher version:
+- Kubernetes version (use `kubectl version`):
+- Cloud provider or hardware configuration:


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,23 @@
+---
+name: Bug report
+about: Tell us about a problem you are experiencing
```
fgiudici, line 3: Nit, I would use "the" here

```suggestion
about: Tell us about the problem you are experiencing
```

.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -0,0 +1,20 @@
+---
+name: Feature enhancement request
+about: Suggest an idea for this project
+
+---
+
+**Describe the solution you'd like**
```
fgiudici, line 7:
```suggestion
**Describe the solution you'd like:**
```

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,23 @@
...
+**What did you expect to happen:**
+
+
+**Anything else you would like to add:**
+[Miscellaneous information that will assist in solving the issue.]
+
+
+**Environment:**
+
+- Elemental version:
```
fgiudici, line 20: What do we expect the user will report here? The version of the elemental operator (the helm chart we release), right? 
Looks good for this PR, probably we may want in the near future (another PR) to review the README of the repo and to update the RELEASES page --> created issue https://github.com/rancher/elemental/issues/363

fgiudici (approved):
a couple of nits, looks good here!

.github/ISSUE_TEMPLATE/bug_report.md
ldevulder: I think in this repo more the `Elemental` version itself (ISO/iPXE), otherwise the draft should be in `elemental-operator`, no? Or also the version of the CLI and the toolkit might be needed (toolkit less interesting in fact maybe)?

fgiudici: Good point, makes sense, the ISO is likely the only thing that we should address here.
What confused me a bit is that while previously there was no operator and the cluster installation logic was managed by the binaries inside the ISO (so basically the ISO was Elemental, a 1:1 match), now there is the operator that takes care of the cluster provisioning part and that manages the configuration for the Elemental installation. So in my mind an "Elemental" release is made up of both the ISO and the operator chart.
Anyway, guess all will be more clear as soon as we update the Release section .

.github/ISSUE_TEMPLATE/bug_report.md
mjura: ups, sorry fixed

.github/ISSUE_TEMPLATE/feature_request.md
mjura: thank you, fixed

.github/ISSUE_TEMPLATE/bug_report.md
mjura: I think we should tell users to put here output from `cat /etc/os-release` 

What do you think ?

davidcassany (approved):
lgtm

.github/ISSUE_TEMPLATE/bug_report.md
ldevulder: I'm fine with the content of `/etc/os-release`, it should have we need, at least to know the image used.

I know that @Itxaka also begin to code an `elemental-support` tool that was/is mainly for dev purposes, but maybe it could be a good idea to add the resulting file in an issue? Let's see this for a later evolution of the draft, as I don't know if this tool will still exists.

PR #359, opened 2022-09-23
Use EFI firmware for CI

### LINKED ISSUE #358: Use EFI boot in integration tests
BIOS boot is currently used in integration tests, we should now use EFI boot instead. We just need to check how to do this in VirtualBox.

ldevulder:
Should fix #358.


fgiudici (approved):
looks good here!

PR #378, opened 2022-09-29
ci/e2e: use EFI firmware instead of deprecated BIOS

ldevulder:
Should fix https://github.com/rancher/elemental/issues/340.

Verification run: https://github.com/ldevulder/elemental/actions/runs/3150802958 (only one node, my lab is not powerful enough to test the whole stack).


davidcassany (approved):
Nice! Looks good :+1:

juadk (approved):

Issue #374, opened 2022-09-28
Cluster creation gets stuck while provisioning 


agracey:
**What steps did you take and what happened:**
Following the quickstart guide. Once I install the OS and reboot, I see the elemental-system-agent (ESA) get preemepted by the rancher-system-agent (RSA) but then the RSA sits there and doesn't install k3s.

At this point, I can see that everything is set up correctly in the upstream CRDs but the MachineInventory is stuck waiting for the plan to be applied. The discrepancy between the applied and plan checksum in the status field shows this to be the case. 

When I log into the host and view the journal for ESA, I see that the plan was indeed applied and that the MachineInventory status is not correct. Since I've seen this before, I guessed that it was because the status didn't get synced before the RSA started and stopped the ESA from running.

A workaround to get my cluster up and running is to run these commands:
- systemctl stop rancher-system-agent
- mv rancher-system-agent.service rancher-system-agent.service.back
- systemctl start elemental-system-agent
- sleep 5
- systemctl stop elemental-system-agent
- mv rancher-system-agent.service.back rancher-system-agent.service
- systemctl start rancher-system-agent

The cluster should start being provisioned after this.

**What did you expect to happen:**
I expect the cluster to be created without manual intervention. 

**Anything else you would like to add:**

I'll be demoing to the community in a master class next Wednesday and would prefer to not need a manual workaround for this.

**Environment:**

- Elemental release version (use `cat /etc/os-release`): 0.6.0
- Rancher version: 2.6.7
- Kubernetes version (use `kubectl version`): upstream  1.24.10+k3s1 -- downstream v1.23.7+k3s1
- Cloud provider or hardware configuration: digital ocean upstream -- Intel NUCs downstream

fgiudici:
rancher-system-agent (RSA)  tears down the elemental-system-agent (ESA) before moving on to the cluster provisioning plan.
In this case as stated above it does it a bit too early, likely before the ESA is able to communicate that the rancher-system-agent deployment plan was successful.
Proper solution is to rework the handover from ESA to RSA.

In the meanwhile, one trick that would help is to wait stopping the elemental-system-agent when the stop command is received.
Adding a section like `ExecStop=sleep 10` to the _elemental-system-agent.service_ in the ISO should make the trick.

I tested adding a long delay (5  minutes) before tearing down the elemental-system-agent. That required to change also the default stop timeout in systemd for the ESA service:
`ExecStop=sleep 300`
`TimeoutStopSec=6min`

[...]

PR #382, opened 2022-09-30
elemental-system-agent: delay the service stop of few seconds

fgiudici:
This is a workaround for #374: when stopping the elemental-system-agent service, we will wait 10 seconds to tear down the service (which will likely be performed by the newly spawned rancher-system-agent). Let's elemental-system-agent have a bit of seconds more to mark completed the rancher-system-agent startup plan.

### DIFF
--- a/framework/files/usr/lib/systemd/system/elemental-system-agent.service
+++ b/framework/files/usr/lib/systemd/system/elemental-system-agent.service
@@ -15,3 +15,4 @@ StandardError=journal+console
 Environment="CATTLE_AGENT_CONFIG=/etc/rancher/elemental/agent/config.yaml"
 Environment="CATTLE_LOGLEVEL=debug"
 ExecStart=/usr/sbin/elemental-system-agent sentinel
+ExecStop=sleep 10


framework/files/usr/lib/systemd/system/elemental-system-agent.service
```diff
@@ -15,3 +15,4 @@ StandardError=journal+console
 Environment="CATTLE_AGENT_CONFIG=/etc/rancher/elemental/agent/config.yaml"
 Environment="CATTLE_LOGLEVEL=debug"
 ExecStart=/usr/sbin/elemental-system-agent sentinel
+ExecStop=sleep 10
```
Itxaka, line 18: This wont stop the agent from runnig, will it? This just substitutes the command that systemd will call to stop the service IIRC

Itxaka: disregard this, manual says otherwise.

Itxaka (approved):

davidcassany (approved):
OK I am fine having it at 10 seconds, lets hope it will be enough for now. In any case we should leave the former issue opened just to find/discuss better options.

PR #395, opened 2022-10-04
ci/e2e: increase test to 5 nodes

ldevulder:
Asked by Paul ;-)

**Note:** I'm not sure that the runner is powerful enough to handle 5 nodes, but it can be easily increased. So let's see how the first run with 5 nodes goes.


juadk (approved):
LGTM

PR #401, opened 2022-10-06
Use a dedicated runner for UI test (spot VM)

### LINKED ISSUE #398: Add a new runner for the UI tests using the spot feature
After a short period of observation, it's difficult to handle all the tests with only one runner in the Cloud. We have to manage locks, pending state, jobs are killed when new one arrives etc...

Moreover we would like to migrate to spot instances to reduce the bill by using the google-preemptible flag
```
This reduces the costs for an instance dramatically (usually the prementible instance costs ~33% of the normal instance) with the disadvantage, that the instance might be terminated by Google. In this (rather rare) case the job fails, and you need to retry it.
```

So, we want to create another runner using the spot feature, in a firstime, it will be dedicated to the UI tests but at the end, we would like to share both runners for backend and UI tests.

juadk:
Fix #398 

Verification run:
https://github.com/rancher/elemental/actions/runs/3198051962
This test failed but for a reason (the sporadic issue I got) external to this PR, we can see that the job was executed on the new runner.

This PR changes the runner used for UI testing, actually it is dedicated to UI.
The instance is configured as a spot instance to save money (153$ per month instead of 311$)

### DIFF
--- a/.github/workflows/ui-e2e.yaml
+++ b/.github/workflows/ui-e2e.yaml
@@ -19,7 +19,7 @@ jobs:
       cluster_name: cluster-k3s
       dashboard_version: elemental-dev
       k8s_version_to_provision: v1.23.10+k3s1
-      runner: elemental-e2e-ci-runner-x86-64-1
+      runner: elemental-e2e-ci-runner-spot-x86-64-1
       start_condition: success
       test_type: ui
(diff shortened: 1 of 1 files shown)


mjura (approved):
LGTM

PR #414, opened 2022-10-11
Add script to inject registration yaml into ISO

### LINKED ISSUE #376: Adapt our elemental-build-iso script to consume released ISOs
This card is about adapting elemental-build-iso script to actually run a container that simply fetches a released ISO and adds the registration yaml in it.

As stated in #365 this procedure requires Xorriso 1.5.4 (currently on TW). A consequence of this might be considering shipping our own Xorriso package meanwhile we attempt to get it updated in the OS code base.

Follow up of #365
Requires #375

davidcassany:
This commit adds a script that downloads the ISO from Dev, Staging or Stable projects and the associated builder image and then injects a given registration yaml file into the ISO.

Results with a new ISO into the current directory.

Fixes #376

### DIFF
--- a/.github/elemental-iso-add-registration
+++ b/.github/elemental-iso-add-registration
@@ -2,8 +2,25 @@
 
 set -e
 
+usage() {
+me=$(basename "$0")
+cat << EOF
+
+usage: 
+   $me REGISTRATION_CONFIG_FILE [ISO_URL]
+
+REGISTRATION_CONFIG_FILE is the yaml file required by elemental-register in
+order to self register against the elemental-operator.
+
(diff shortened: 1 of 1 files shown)


davidcassany:
This is a first iteration of the script. I have included it under `.github` together with the `elemental-iso-build`, I doubt this is the right place and I am also unsure of what to do with such a script. 

Probably we could also include into the github release and ISO and then also publish this script and include the released ISO default URL? seams a bit convoluted and still we need a builder image including xorriso v1.5.4, currently build only in OBS.

fgiudici (approved):
Looks good here!
Love to not have to rebuild the iso from scratch anymore 🎉

ldevulder (approved):

Issue #412, opened 2022-10-11
ci/e2e: add a test for elemental-operator upgrade


ldevulder:
Currently the latest released version of `elemental-operator` is installed. An upgrade to the latest build for CI version could be done easily to validate the upgrade process.

Also add an option to test the version of `elemental-operator` built in OBS, use this `helm` command to do this:
```
helm upgrade --create-namespace -n cattle-elemental-system --install elemental-operator oci://registry.opensuse.org/isv/rancher/elemental/dev/charts/elemental/elemental-operator
```

PR #426, opened 2022-10-17
e2e: add option to select Rancher Manager version and Teal ISO to test

ldevulder:
Verification run: https://github.com/ldevulder/elemental/actions/runs/3266173805

### DIFF
--- a/tests/e2e/install_test.go
+++ b/tests/e2e/install_test.go
@@ -81,7 +81,7 @@ var _ = Describe("E2E - Install Rancher Manager", Label("install"), func() {
 			err = kubectl.RunHelmBinaryWithCustomErr("repo", "update")
 			Expect(err).To(Not(HaveOccurred()))
 
-			err = kubectl.RunHelmBinaryWithCustomErr("install", "cert-manager", "jetstack/cert-manager",
+			err = kubectl.RunHelmBinaryWithCustomErr("upgrade", "--install", "cert-manager", "jetstack/cert-manager",
 				"--namespace", "cert-manager",
 				"--create-namespace",
 				"--set", "installCRDs=true",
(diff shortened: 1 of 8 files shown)


davidcassany (approved):
Good! Nice refactor to make ISO parametrizable :+1:

For completeness we should also consider elemental-operator helm chart source parametrizable. But we can do that in follow up PRs.

tests/e2e/install_test.go
```diff
@@ -143,7 +146,7 @@ var _ = Describe("E2E - Install Rancher Manager", Label("install"), func() {
 			err = kubectl.RunHelmBinaryWithCustomErr("repo", "update")
 			Expect(err).To(Not(HaveOccurred()))
 
-			err = kubectl.RunHelmBinaryWithCustomErr("install", "elemental-operator", "elemental-operator/elemental-operator",
+			err = kubectl.RunHelmBinaryWithCustomErr("upgrade", "--install", "elemental-operator", "elemental-operator/elemental-operator",
```
davidcassany, line 149: This will install the elemental-operator helm chart and elemental-operator image from the github builds in any case, however it would be nice to that this is pulled form github or OBS depending on test. This turns to be slightly tricky as in OBS there is no helm chart repository published, the helm chart lives in an oci image. 

I install from OBS helm chart with
```
helm upgrade --create-namespace -n cattle-elemental-system --install elemental-operator oci://registry.opensuse.org/isv/rancher/elemental/dev/charts/elemental/elemental-operator
```

So it requires little refactor to adapt this code to make elemental-operator installation source parametrizable. In any case I believe this can be done in follow up PRs :wink:

ldevulder: Yes, thx for the command line.

I would like to pair this change with this issue: https://github.com/rancher/elemental/issues/412.

davidcassany: yes, makes complete sense :+1:

ldevulder:
@davidcassany do you think that these workflow (dev and staging) should be nightly scheduled? Or dispatch is enough?

davidcassany:
> @davidcassany do you think that these workflows (dev and staging) should be nightly scheduled? Or dispatch is enough?

I'd say dispatch is enough for now

PR #443, opened 2022-10-21
docs: update links

fgiudici:
We now have the 'staging' repo which builds elemental images when we tag (release) the elemental-operator repo: update the links.


alexander-demichev (approved):

kkaempf:
Uhm. no quite what we actually wanted.

`staging` is the preparation area before we release. This is what QA is using for running acceptance tests. Once all tests pass, the sofware should be promoted to `stable`.

Since `stable` is still missing, documenting `staging` is a workaround but not acceptable for release.

/cc @davidcassany

PR #450, opened 2022-10-24
ci/e2e: add retry on curl command

ldevulder:
To avoid sporadic failure like this one: https://github.com/rancher/elemental/actions/runs/3312488264/jobs/5477683940

Tested locally, I'm not sure that this will fix the issue TBH, but at least there's no regression.


kkaempf:
Please also add a `-v` to the `curl` command and save the output. I'd guess we will see some redirects and evtl. problematic mirrors 🤔

kkaempf (approved):

PR #460, opened 2022-10-25
Network: Make NM ignore calico and flannel devices

### LINKED ISSUE #446: [rke2] configure NetworkManager to ignore calico/flannel related network interfaces.
According to rke2 docs we need to tell NM to not manage the calico/flannel interfaces in case we are deploying rke2 with elemental.

This is in the middle of being a bug and a feature to be honest, I havent find any issues when deploying rke2 but its tru that we were using wicked back then and now elemental ships with NM by default.

https://docs.rke2.io/known_issues/#networkmanager

Itxaka:
According to rke2 docs NM can interfere with CNI routing.

This patch adds a config file so NM ignores devices from both calico and flannel

Fixes #446 


kkaempf (approved):
:+1:

PR #463, opened 2022-10-25
ci: Fix github deprecations

Itxaka:

### DIFF
--- a/.github/workflows/integration.yaml
+++ b/.github/workflows/integration.yaml
@@ -29,7 +29,7 @@ jobs:
       - name: Install dependencies
         run: brew install cdrtools jq
       - name: Wait for build
-        uses: fountainhead/action-wait-for-check@v1.0.0
+        uses: fountainhead/action-wait-for-check@v1.1.0
         id: wait-for-build
         with:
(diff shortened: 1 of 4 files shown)


davidcassany (approved):
Changes look to me, I just don't see where this netlify checks are comming from and what are they supposed to mean/check...

.github/workflows/integration.yaml
```diff
@@ -29,15 +29,15 @@ jobs:
...
+        uses: fountainhead/action-wait-for-check@v1.1.0
         id: wait-for-build
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
           checkName: build-all-artifacts
           ref: ${{ github.event.pull_request.head.sha || github.sha }}
           timeoutSeconds: 1800
       - name: Download artifacts
-        uses: dawidd6/action-download-artifact@v2
+        uses: dawidd6/action-download-artifact@v2.24.0
```
ldevulder, line 40: Not sure that this change was needed, as far as I understand `dawidd6/action-download-artifact@v2` should point to the latest v2 version, no?

PR #468, opened 2022-10-26
Use the elemental plugin in the UI CI

### LINKED ISSUE #445: Test and document the new plugin way to enable elemental UI
There is a new way to deploy the elemental UI on top of Rancher:
image

image

- [x] Test it manually
- [x] Write private documentation
- [x] Check if the opened bugs are fixed
- [x] Use it in the CI

Requirements:
rancher v2.7.0-rc4
elemental-operator
elemental-ui repo

juadk:
Fix #445 

The new plugin/extension is ready to test so let's integrate in in the CI instead of using a tricky older dev branch.


cypress/support/elemental.ts
```diff
@@ -24,15 +24,13 @@ export class Elemental {
...
+    expect($lis).to.have.length(6);
     expect($lis.eq(0)).to.contain('Dashboard');
     expect($lis.eq(1)).to.contain('Machine Registrations');
     expect($lis.eq(2)).to.contain('Machine Inventories');
-    expect($lis.eq(3)).to.contain('Mach. Inv. Selectors');
-    expect($lis.eq(4)).to.contain('Mach. Inv. Selec. Templates');
-    expect($lis.eq(5)).to.contain('Managed OS Versions');
-    expect($lis.eq(6)).to.contain('Managed OS Version Channels');
-    expect($lis.eq(7)).to.contain('OS Image Upgrades');
+    expect($lis.eq(3)).to.contain('Managed OS Versions');
```
kkaempf, line 31: Alexandre talked about dropping these fields as their semantics isn't fully defined yet.

kkaempf (approved):
lgtm

package.json
```diff
@@ -1,7 +1,7 @@
 {
-  "name": "epinio-end-to-end-tests",
+  "name": "elemental-ui-e2e",
```
Itxaka, line 2: :eyes:

tests/e2e/install_test.go
```diff
@@ -128,7 +128,11 @@ var _ = Describe("E2E - Install Rancher Manager", Label("install"), func() {
 
 			// Set specified version if needed
 			if rancherVersion != "" && rancherVersion != "latest" {
-				flags = append(flags, "--version", rancherVersion)
+				if rancherVersion == "devel" {
+					flags = append(flags, "--devel")
```
Itxaka, line 132: I dont think this are exclusive no? You can set --devel AND a version, only --devel will inlcude in the list all the devel versions.

Maybe it will be useful if the latest devel is broken but we need to test with the previous devel version?

Itxaka (approved):

cypress/support/elemental.ts
juadk: yep exactly

tests/e2e/install_test.go
ldevulder: `--devel` isn't needed if you specified a version. We tried this morning and you can set `--version 2.7.0.-rc7` and it will install this development/RC version without the need of `--devel`. `--devel` is useful here to force the installation of the latest development version.

So with the variables in the workflow files you can set this if you always want the latest version (which is needed for UI):
```
with:
  rancher_channel: latest
  rancher_version: devel
```

And this if you want a specific development version:
```
with:
  rancher_channel: latest
  rancher_version: 2.7.0-rc4
```

Issue #447, opened 2022-10-24
Elemental backup and restore tests


mjura:

mjura:
It seems that it is this same problem https://github.com/rancher/rancher/issues/34107

PR #481, opened 2022-11-01
Add warning to restore procedure about rancher-system-agent

mjura:
Fixes https://github.com/rancher/elemental/issues/447

### DIFF
--- a/docs/restore.md
+++ b/docs/restore.md
@@ -46,3 +46,32 @@ INFO[2022/10/31 06:34:50] Done restoring
 ```
 
 Continue with procedure from [Rancher documentation](https://docs.ranchermanager.rancher.io/how-to-guides/new-user-guides/backup-restore-and-disaster-recovery/migrate-rancher-to-new-cluster)
+
+:::warning warning
+Try to not restart `rancher-system-agent` on elemental nodes, otherwise it can stuck in following error
+
+`panic: error while connecting to Kubernetes cluster: the server has asked for the client to provide credentials`
+
+If this problem occurs please go with procedure below.
+:::
+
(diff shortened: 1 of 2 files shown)


docs/restore.md
```diff
@@ -46,3 +46,32 @@ INFO[2022/10/31 06:34:50] Done restoring
 ```
 
 Continue with procedure from [Rancher documentation](https://docs.ranchermanager.rancher.io/how-to-guides/new-user-guides/backup-restore-and-disaster-recovery/migrate-rancher-to-new-cluster)
+
+:::warning warning
+Try to not restart `rancher-system-agent` on elemental nodes, otherwise it can stuck in following error
+
+`panic: error while connecting to Kubernetes cluster: the server has asked for the client to provide credentials`
+
+If this problem occurs please go with procedure below.
```
Itxaka, line 55: I wonder if we should just start a troubleshooting page and start putting this info in a converged place.

scripts/rancher-agent-token-update
```diff
@@ -0,0 +1,47 @@
+#!/bin/sh
+
+if [ "${DEBUG}" = 1 ]; then
```
Itxaka, line 3: Maybe we should include this functionality in the register client? That is expected to be found in the node and it's related to teal....

Itxaka (approved):

scripts/rancher-agent-token-update
mjura: it is my plan for next research, I would like to move into register client

mjura: So problem is that after restoring Rancher from backup all secrets will be regenerated and even for elemental register client this will change too, only one way to update it is to run provided script `rancher-agent-token-update` on node where we have kubeconfig for Kubernetes Rancher cluster.

PR #485, opened 2022-11-02
Add RBAC tests in the UI

### LINKED ISSUE #385: UI - Test RBAC role
I think we need to test RBAC role in the UI.
https://github.com/rancher/elemental/issues/332

juadk:
image

Regarding RBAC tests, I don't know how deep they have to be, I mean, with the existing tests, I only test that a user with the elemental-administrator can access the OS management menu. I also test the opposite, a user without the elemental-administrator role should not access the OS management menu.

By deeper, I mean, we can test that a user with the elemental-administrator role can create machine registration etc
May I ask what do you think @kkaempf @ldevulder

### DIFF
--- a/.github/workflows/master-e2e.yaml
+++ b/.github/workflows/master-e2e.yaml
@@ -162,6 +162,7 @@ jobs:
             cypress/e2e/unit_tests/elemental_plugin.spec.ts
             cypress/e2e/unit_tests/menu.spec.ts
             cypress/e2e/unit_tests/machine_registration.spec.ts
+            cypress/e2e/unit_tests/user.spec.ts
         run: make -f tests/Makefile start-cypress-tests
       - name: Deploy a node to join Rancher manager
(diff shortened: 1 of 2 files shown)


ldevulder:
> May I ask what do you think @kkaempf @ldevulder

One thing we can do (if it's possible) will be to test for example K3s cluster deployment with `admin` user and RKE2 with a user+elemental-administrator role?

juadk:
> > May I ask what do you think @kkaempf @ldevulder
> 
> One thing we can do (if it's possible) will be to test for example K3s cluster deployment with `admin` user and RKE2 with a user+elemental-administrator role?

I think that's possible to implement this solution very quickly once the bug fixed.

mjura (approved):
LGTM

juadk:
Ok, let's merge this one and I will submit another PR to use a normal user in the RKE2 test case once the fix provided by the UI team.

PR #486, opened 2022-11-02
Update persistent paths for NetworkManager

frelon:
SLE Micro 5.3 replaces wicked with NetworkManager.

https://www.suse.com/releasenotes/x86_64/SLE-Micro/5.3/index.html


kkaempf (approved):
Good catch !

davidcassany (approved):

PR #493, opened 2022-11-03
First take on reflow

agracey:


docs/quickstart.md
```diff
@@ -15,6 +15,16 @@ with the only help of an Elemental Teal iso
 
 ---
 
+### What is the Rancher Elemental Stack ?
+
+The Elemental Stack consists of some packages on top of SLE Micro for Rancher
+
+- **elemental-toolkit** - includes a set of OS utilities to enable OS management via containers. Includes dracut modules, bootloader configuration, cloud-init style configuration services, etc.
+- **elemental-operator** - this connects to Rancher Manager and handles machineRegistration and machineInventory CRDs
```
kkaempf, line 23: elemental-operator is not "on top of SLE Micro for Rancher" and should be listed separately

docs/quickstart.md
```diff
@@ -15,6 +15,16 @@ with the only help of an Elemental Teal iso
...
+- **elemental-register** - this registers machines via machineRegistrations and installs them via elemental-cli
+- **elemental-cli** - this installs any elemental-toolkit based derivative. Basically an installer based on our A/B install and upgrade system
+- **rancher-system-agent** - runs on the installed system and gets instructions ("Plans") from Rancher Manager what to install and run on the system
```
kkaempf, line 26: we all wished that things would be _that_ simple.
In reality, we run `elemental-system-agent` first which only purpose is a single plan: Download an start `rancher-system-agent` 🤦🏻‍♂️

docs/quickstart.md
```diff
@@ -26,31 +36,24 @@ Its sole purpose is to run Kubernetes (k3s or RKE2), with everything controlled
...
+     - The Rancher server needs to be accessible on port 443 from all nodes that you are provisioning. (However, the requests are only one way so the nodes can be behind a NAT or other firewall)
  - A machine (bare metal or virtualized) with TPM 2.0
      - Hint 1: Libvirt allows setting virtual TPMs for virtual machines [example here](https://rancher.github.io/elemental/tpm/#add-tpm-module-to-virtual-machine)
-     - Hint 2: You can enable TPM emulation on bare metal machines missing the TPM 2.0 module [example here](https://rancher.github.io/elemental/tpm/#add-tpm-emulation-to-bare-metal-machine)
+     - Hint 2: You can enable TPM emulation on bare metal machines missing the TPM 2.0 module [example here](https://rancher.github.io/elemental/tpm/#add-tpm-emulation-to-bare-metal-machine). There are many caveats to this but it's useful for trying out with a single node. 
  - Helm Package Manager (https://helm.sh/)
  - Docker (for iso manipulation)
 
-## Preparing the cluster
+## Preparing the upstream cluster
```
kkaempf, line 50: "management" cluster ? 🤔

docs/quickstart.md
```diff
@@ -243,6 +236,72 @@ This will generate an ISO on the current directory with the name `elemental-<tim
 </TabItem>
 </Tabs>
 
+
+Note: Much but not all of the configuration provided in a `MachineRegistration` will be read at installation time instead of baked into the installation media. This means that many changes will not require using a new ISO.
```
kkaempf, line 240: This is still somewhat fuzzy

The only purpose of the embedded-in-the-ISO configuration is to establish the network connection to the management cluster. Everything else will be loaded from this cluster.

docs/quickstart.md
```diff
@@ -243,6 +236,72 @@ This will generate an ISO on the current directory with the name `elemental-<tim
...
+
+
+## Boot your nodes
+You can now boot your nodes with this newly created ISO, and they will:
+
+ - Register with the registrationURL given and create a per-machine `MachineInventory`
```
kkaempf, line 246: ... using the TPM hash as the unique machine identifier.

docs/quickstart.md
```diff
@@ -243,6 +236,72 @@ This will generate an ISO on the current directory with the name `elemental-<tim
...
+ - Install Elemental Teal to the given device
+ - Shutdown
```
kkaempf, line 248: This needs more explanation as the shutdown-or-reboot behavior can be controlled in the machineRegistration

docs/quickstart.md
```diff
@@ -243,6 +236,72 @@ This will generate an ISO on the current directory with the name `elemental-<tim
...
+ - Call home to your Rancher and wait
```
kkaempf, line 249: Huh ? After the shutdown ? 😉

kkaempf, line 249: Huh ? After the shutdown ? 😉

docs/quickstart.md
```diff
@@ -243,6 +236,72 @@ This will generate an ISO on the current directory with the name `elemental-<tim
...
+
+After a few minutes you will see your nodes show up in the `MachineInventory` list.
```
kkaempf, line 251: The machineInventory entry will be created as soon as elemental-register makes the first contact to the management cluster. That doesn't need "a few minutes" but happens rather immediately after ISO boot.

docs/quickstart.md
```diff
@@ -243,6 +236,72 @@ This will generate an ISO on the current directory with the name `elemental-<tim
...
+
+## Build a Cluster
+
+Lastly, we need to build clusters out of our nodes! 
+
+This step can be done at any time (including before the machines were installed)
+
+We need two resources per cluster: a `Cluster` resource and a `MachineInventorySelectorTemplate` to know which machines are contained in the cluster.
+
+This selector is a simple matcher based on labels set in the `MachineInventory`, so if your selector is matching the `cluster-id` key with a value `myId`
```
kkaempf, line 267: Using labels is one possibility for the matcher. There are other matchers available.

kkaempf (changes requested):
:+1: on the first take

I added some remarks :wink:

docs/quickstart.md
mjura: I agree, that is not related to SLE Micro, but it would great anyway have listed it here on this `quickstart` page

docs/quickstart.md
```diff
@@ -26,31 +36,24 @@ Its sole purpose is to run Kubernetes (k3s or RKE2), with everything controlled
...
-- **elemental-operator** - this connects to Rancher Manager and handles machineRegistration and machineInventory CRDs
-- **elemental-register** - this registers machines via machineRegistrations and installs them via elemental-cli
-- **elemental-cli** - this installs any elemental-toolkit based derivative. Basically an installer based on our A/B install and upgrade system
-- **rancher-system-agent** - runs on the installed system and gets instructions ("Plans") from Rancher Manager what to install and run on the system
-
 ## Prerequisites
 
  - A Rancher server (2.6.9) configured (server-url set)
      - To configure the Rancher server-url please check the [Rancher docs](https://rancher.com/docs/rancher/v2.6/en/admin-settings/#first-log-in)
+     - The Rancher server needs to be accessible on port 443 from all nodes that you are provisioning. (However, the requests are only one way so the nodes can be behind a NAT or other firewall)
```
mjura, line 43: In case if Rancher can be exposed with public domain then `ngrok` project can be used.

docs/quickstart.md
mjura: All K8s clusters managed by Rancher are called `downstream` clusters https://docs.ranchermanager.rancher.io/reference-guides/rancher-manager-architecture/communicating-with-downstream-user-clusters

docs/quickstart.md
agracey: Ok, I'll drop this line. I added it to suggest that you wouldn't need to rebuild (or redownload) the iso after every change.

docs/quickstart.md
agracey: I think I recopied a prior change 😆

docs/quickstart.md
agracey: Added a note on where to find more options

docs/quickstart.md
agracey: 👍

docs/quickstart.md
agracey: Changed to `elemental-system-agent`. As this is the quickstart, I'm hesitant to make the distinction here.

docs/quickstart.md
agracey: I gave a little bit more context instead of "on top of SLE Micro". I'm not in love with what I wrote but it's early on a monday.

docs/quickstart.md
```diff
@@ -223,26 +220,73 @@ Now that we have the script we can proceed to download the ISO and inject our co
 
 This will generate an ISO on the current directory with the name `elemental-teal-<ARCH>.iso`
 
-:::info
-The script uses the iso for the arch based on the system is being run from. If you want to cross build for another system,
-you can set the `ARCH` environment variable to the desired target system (x86_64, aarch64) and the iso will be build for that architecture.
-:::
+</TabItem>
+</Tabs>
```
Itxaka, line 224: This just dropped the create iso script. Do we want to drop it completely? Or do we want to keep it and make clear that there is 2 methods to create the iso via script? IMO, I would drop it fully, use only the inject one AND add an extra tab to explain how to build the iso FROM the repo, making sure we explain that is only meant for DEVELOPMENT, trying new stuff, etc...

Or even create a new section for that? While I think we should point as much thinhgs as possible to use th official OBS images, I think its important to explain how you can build your own iso from scratch and add stuff on it, test new things and so on, so building from the repo with `make build && make build_iso`

Yeah, maybe an extra section for building iso or developments is needed to explain that.

agracey: I think it's fine to drop it from the quickstart. It might fit better in the "customize the OS" page?

kkaempf:
Not sure why this never got merged. 😞 

Docs have move to rancher/elemental-docs and have a different structure meanwhile.

PR #502, opened 2022-11-08
docs: Use variables for operator name

frelon:
Not sure where the link should point, but hopefully this is correct :+1: 


nunix (approved):
Good catch, thank you.

PR #507, opened 2022-11-09
Use machine reg created in UI instead of the one in test assets

### LINKED ISSUE #215: UI - Finish the custom cloud config test in machine registration
This one should have been done in https://github.com/rancher/elemental/issues/196 but a [bug](https://github.com/rancher/elemental-ui/issues/15) was discovered regarding cloud config.

Once the bug fixed, we will be able to finish this test.

juadk:
Due to a bug with cloud-config, until now, we were using a machine registration not created in the UI but the same as the backend e2e test (https://github.com/rancher/elemental/blob/main/tests/assets/machineregistration.yaml).

It sound more logical to run the test with a machine registration created with the UI.

This PR adds a test in machine registration to create the resource we need to deploy a node.

Actually, it also tests the cloud config block because we have to import our custom cloud config there.
Fix #215 

image

Verification run:
https://github.com/rancher/elemental/actions/runs/3427858809

### DIFF
--- a/.github/workflows/master-e2e.yaml
+++ b/.github/workflows/master-e2e.yaml
@@ -292,7 +292,7 @@ jobs:
         if: always()
         run: |
           # Remove all VMs
-          for I in $(sudo virsh list | awk '/node-/ { print $2 }'); do
+          for I in $(sudo virsh list --all| awk '/node-/ { print $2 }'); do
             for C in destroy undefine; do
               [[ "${C}" == "undefine" ]] && OPT="--nvram" || unset OPT
               sudo virsh ${C} ${OPT} ${I} >/dev/null 2>&1 || true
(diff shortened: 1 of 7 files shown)


.github/workflows/master-e2e.yaml
```diff
@@ -292,7 +292,7 @@ jobs:
         if: always()
         run: |
           # Remove all VMs
-          for I in $(sudo virsh list | awk '/node-/ { print $2 }'); do
+          for I in $(sudo virsh list -all| awk '/node-/ { print $2 }'); do
```
ldevulder, line 295: should be `--all`, not `-all`.

ldevulder (changes requested):

ldevulder (approved):

PR #508, opened 2022-11-09
docs: Change nodeSelector example

frelon:
The smbios labels are not propagated to the nodes, so `location` cannot be used for nodeSelector in any meaningful way.

### DIFF
--- a/examples/upgrade/upgrade-node-selector.yaml
+++ b/examples/upgrade/upgrade-node-selector.yaml
@@ -10,4 +10,4 @@ spec:
     - clusterName: my-cluster
   nodeSelector:
     matchLabels:
-      location: "europe"
+      kubernetes.io/hostname: my-machine


examples/upgrade/upgrade-node-selector.yaml
```diff
@@ -10,4 +10,4 @@ spec:
     - clusterName: my-cluster
   nodeSelector:
     matchLabels:
-      location: "europe"
+      kubernetes.io/hostname: my-machine
```
Itxaka, line 13: This is part of the example done on the quickstart and the node-selector stuff. i.e. This shows how to put some labels on your registration and then use the nodeselector with that specific label to select those: https://github.com/rancher/elemental/blob/main/examples/quickstart/registration.yaml#L19

frelon: Exactly, but those labels are not propagated to the nodes, so trying to manage upgrades with this example is not going to work as expected right now (as discussed in #509)

Itxaka: wait, really? I can see in a freshly deployed node that the machineinventory has the labels set properly according to the registration using the default quickstart yamls?

```
$ kubectl describe machineinventory --all-namespaces
Name:         my-machine
Namespace:    fleet-default
Labels:       location=europe
              machineUUID=c6e10c11-4655-49ad-a10a-ddaa66e4c630
              manufacturer=QEMU
              productName=Standard-PC-Q35-ICH9-2009
              serialNumber=Not-Specified
```

Itxaka: oh I SEE to the actual NODE! interesting!

Itxaka (approved):

PR #510, opened 2022-11-09
Added Algolia search bar

nunix:
The Algolia search crawler has been configured and the search bar can now be used.

Another small change to the configuration was also done for the URL and baseURL settings to avoid a warning when building the docs website.

Signed-off-by: Nuno do Carmo <nuno.carmo@suse.com>


kkaempf (approved):
ack

PR #515, opened 2022-11-11
Add a 'Slack' link next to 'GitHub'

kkaempf:


Itxaka (approved):
We should add icons for both :D

kkaempf:
> We should add icons for both :D

:+1:
not sure how to do this, https://docusaurus.io/docs/api/themes/configuration/#navbar-items isn't really conclusive on this 😞

Issue #522, opened 2022-11-16
Fully automated rebuilds on tags


davidcassany:
Automated builds on tags in staging are not happening due to a bug in OBS related to SCM/CI integration.

Tracker for https://github.com/openSUSE/open-build-service/issues/13206

kkaempf:
Merged upstream.

davidcassany:
Verification failed. Upstream issue in OBS has been reopened.

PR #523, opened 2022-11-16
Also trigger Dev rebuild on tag push

davidcassany:
This commit sets the OBS workflow to also rebuild Dev environment when a tag is pushed to main branch. This is relevant to ensure the latest Dev build is also taking the correct package version.

### DIFF
--- a/.obs/workflows.yml
+++ b/.obs/workflows.yml
@@ -1,4 +1,4 @@
-dev_workflow:
+push_workflow:
   steps:
     - trigger_services:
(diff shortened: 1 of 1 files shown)


davidcassany:
Unfortunately due to #522 this can be verified (that's mostly the reason I did not commit it yet). Hopefully OBS will be fixed soon, the bug is on their planned column in their public project board 🤞🏽

Itxaka (approved):

PR #525, opened 2022-11-16
Do not overwrite VERSION/VERSION_ID on dockerfile build

Itxaka:
We are overriding the existing values in the os-release file and we
should not. We already have the IMAGE_TAG and/or COMMIT to check the
versions

Also adds the TIMESTAMP to os-release

Both are done to be more in sync with OBS builds.


frelon (approved):
I'm not very familiar with OBS, but sounds good to me!

PR #527, opened 2022-11-16
ci/e2e: make Rancher version configurable

ldevulder:
Also set `latest/devel` version as the default one (but this can be changed upon triggered).

Verification run: https://github.com/rancher/elemental/actions/runs/3482867640


juadk (approved):
LGTM

PR #533, opened 2022-11-21
don't reference outdated image projects

kkaempf:
...like SLE Micro for Rancher 5.2


Itxaka (commented):
unfortunately we may need to keep this updated anytime we bump unless we move it to a different image (github ci image?)

Itxaka (approved):

PR #544, opened 2022-11-29
ci/e2e: use new elemental-operator helm chart

ldevulder:
The new operator URI is now aligned with other Rancher charts. This PR also allows the value to be overwritten.


davidcassany (approved):
Awesome, thanks! :+1:

kkaempf (commented):
👍🏻

PR #547, opened 2022-11-29
Drop machineName from registration.yaml example

kkaempf:
to prevent duplicate hostnames in a cluster


Itxaka:
this is already in https://github.com/rancher/elemental-docs/pull/8 

Should we merge it here or we just drop the docs in here and call it a day?

Itxaka (changes requested):
registration-tpm.yaml also has this

kkaempf:
elemental.docs.rancher.com is online (with this fix) 🎉

PR #550, opened 2022-12-01
Set serial console baudrate to 115200

kkaempf:
This is the standard baudrate on Raspberry Pi.


ldevulder:
Just small comments:
- without any value it should be auto detected
- also Serial-Over-Lan on IPMI is be default 19200, so this might have an impact on some hardware.

kkaempf:
> Just small comments:
> 
>     * without any value it should be auto detected

Doesn't work on RPi 😞 

> 
>     * also Serial-Over-Lan on IPMI is be default 19200, so this might have an impact on some hardware.

Hmm, rancher/elemental-cli#297 😉 

Maybe restrict this change to `aarch64` for now ? 🤔

Itxaka (changes requested):
dont really like this. Is RPI a supported platform? We are potentially breaking all the rest of the stuff to cater to a small rpi niche :/

Cant this be fixed in a different way? is the problem on raw image generation? Maybe we can add some overrides or flags to the raw image generation instead, leave them by default and allow to override it somehow= 

How has this appear now? cOS has been tested a lot on RPI IIRC with no issues with this?

kkaempf:
> dont really like this. Is RPI a supported platform?

Not supported, but sufficiently easy to use with Elemental. It also makes a good demo (and we have plenty of requests for it) 😉 

> We are potentially breaking all the rest of the stuff to cater to a small rpi niche :/

That's bad, I agree.

> 
> Cant this be fixed in a different way? is the problem on raw image generation? Maybe we can add some overrides or flags to the raw image generation instead, leave them by default and allow to override it somehow=

Yes, that would be awesome. Some kind of `platform` flag/parameter ...

> 
> How has this appear now? cOS has been tested a lot on RPI IIRC with no issues with this?

Because it's not an issue with cOS/Elemental as such but when testing/debugging. One usually doesn't have a serial console attached. But when you have, you want the Kernel to use the same baudrate like the firmware/bootloader.

kkaempf:
Closing this PR to find a more flexible approach

PR #551, opened 2022-12-02
Added redirect to new URL on all pages

nunix:
Redirect configuration:
from: rancher.github.io/elemental/*
to: elemental.docs.rancher.com

Signed-off-by: Nuno do Carmo <nuno.carmo@suse.com>

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -3,6 +3,13 @@ sidebar_label: Architecture
 title: ''
 ---
 
+<head>
+  <meta charset="utf-8" />
+  <title>Redirecting to https://elemental.docs.rancher.com</title>
+  <meta http-equiv="refresh" content="0; URL=https://elemental.docs.rancher.com/" />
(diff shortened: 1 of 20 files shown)


nunix:
> 20 files changed ? Really ?
> 
> There should be a simpler way ...
> 
> imho we should just redirect the toplevel like we did it for [os2](https://github.com/rancher/os2/pull/20/files)

I wanted to do it too, however while mkdocs allows it on its config, Docusaurus doesn't. So right now, to ensure "all" pages redirect to the new URL, this is the only way, other else we can replace the /docs content with only an index file that does the redirect, but if someone had a page in favorite, then it will throw a 404.

Also, this is an intermediate step and in few months, we'll remove the /docs directory from this repo (to keep it clean).

kkaempf (approved):
Oh well ...

alexander-demicev (approved):
thanks!

PR #563, opened 2022-12-06
ci/ui: fix empty variables in scheduled test

ldevulder:
This should fix this issue: https://github.com/rancher/elemental/actions/runs/3626461866/jobs/6115479156


juadk (approved):
LGTM

PR #574, opened 2022-12-09
Collect elemental support in e2e UI CI

### LINKED ISSUE #417: Collect all logs for e2e tests
When a test fail, all the logs should be collected and uploaded at the end of the test... it will save hours of debugging like it happened in this issue https://github.com/rancher/elemental/issues/410....

From the elemental node:
elemental-support
system-upgrade-controller

From the Rancher node (where Rancher manager is installed):
rancher pod logs
Maybe there is a command to generate all logs from rancher.

juadk:
Fix #417 

Verification Run:
https://github.com/rancher/elemental/actions/runs/3678271212

### DIFF
--- a/.github/workflows/master-e2e.yaml
+++ b/.github/workflows/master-e2e.yaml
@@ -225,7 +225,9 @@ jobs:
             cypress/e2e/unit_tests/machine_selector.spec.ts
             cypress/e2e/unit_tests/machine_inventory.spec.ts
             cypress/e2e/unit_tests/upgrade.spec.ts
-        run: make -f tests/Makefile start-cypress-tests
+        run: |
+          export OPERATOR_VERSION=$(kubectl get pods -n cattle-elemental-system -l app=elemental-operator -o jsonpath={.items[*].status.containerStatuses[*].image}|awk -F ':' '{print substr($2,0,3)}')
+          make -f tests/Makefile start-cypress-tests
       - name: Upload Cypress screenshots (Advanced)
         if: failure()
         uses: actions/upload-artifact@v3
--- a/.github/workflows/master-e2e.yaml
+++ b/.github/workflows/master-e2e.yaml
@@ -23,6 +23,10 @@ on:
         description: Name of the provisioned cluster
         required: true
         type: string
+      elemental_support_version:
+        description: Elemental support version to use for collecting logs
+        default: "1.0.0"
+        type: string
       iso_to_test:
         description: ISO to test (default built one is empty)
         type: string
@@ -302,9 +306,11 @@ jobs:
         run: sudo virsh list
       - name: Store logs
         if: always()
+        env:
+          ELEMENTAL_SUPPORT_VERSION: ${{ inputs.elemental_support_version}}
         run: |
           cd tests && make e2e-get-logs
-      - name: Upload downstream cluster logs
+      - name: Upload cluster logs
         if: always()
         uses: actions/upload-artifact@v3
         with:
--- a/.github/workflows/master-e2e.yaml
+++ b/.github/workflows/master-e2e.yaml
@@ -392,4 +392,3 @@ jobs:
           gcloud --quiet compute instances delete ${{ needs.create-runner.outputs.runner }} \
             --delete-disks all \
             --zone ${{ inputs.zone }}
-
--- a/.github/workflows/master-e2e.yaml
+++ b/.github/workflows/master-e2e.yaml
@@ -300,6 +300,17 @@ jobs:
       - name: List installed nodes
         if: inputs.test_type == 'cli'
         run: sudo virsh list
+      - name: Store logs
+        if: always()
+        run: |
+          cd tests && make e2e-get-logs
+      - name: Upload downstream cluster logs
+        if: always()
+        uses: actions/upload-artifact@v3
+        with:
+          name: support-logs
+          path: tests/**/logs/*
+          if-no-files-found: ignore
       - name: Add summary
         if: always()
         run: |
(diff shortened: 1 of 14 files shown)


.github/workflows/master-e2e.yaml
```diff
@@ -300,6 +300,17 @@ jobs:
       - name: List installed nodes
         if: inputs.test_type == 'cli'
         run: sudo virsh list
+      - name: Store logs
+        if: always()
+        run: |
+          cd tests && make e2e-get-logs
+      - name: Upload downstream cluster logs
```
Itxaka, line 307: This mentions downstream, but we are getting the upstream logs?

tests/e2e/logs_test.go
```diff
@@ -0,0 +1,47 @@
...
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("E2E - Getting logs node", Label("logs"), func() {
+	It("Get the upstream cluster logs", func() {
+		By("Downloading tools to generate logs", func() {
+			_ = os.Mkdir("logs", 0755)
+			_ = os.Chdir("logs")
+			err := exec.Command("curl", "-L", "https://github.com/rancher/elemental-operator/releases/download/v1.0.0/elemental-support_1.0.0_linux_amd64", "-o", "elemental-support").Run()
```
ldevulder, line 17: Maybe better to add a loop here, as the download/configuration code is pretty much the same for both `elemental-support` and the script.

And IMHO also better to use a variable here that can be configured from the workflow (inputs also?) as it too static here.

tests/e2e/logs_test.go
```diff
@@ -0,0 +1,47 @@
...
+			err := exec.Command("./elemental-support").Run()
+			Expect(err).To(Not(HaveOccurred()))
+			err = exec.Command("sudo", "./rancher2_logs_collector.sh", "-d", "../logs").Run()
+			Expect(err).To(Not(HaveOccurred()))
+		})
+
+		By("Collecting additionals logs with kubectl commands", func() {
+			outcmd, err := exec.Command("kubectl", "get", "bundles", "--all-namespaces").CombinedOutput()
+			if err != nil {
+				fmt.Fprint(GinkgoWriter, err)
```
ldevulder, line 37: `GinkgoWriter` is already used in other files but with the form: `GinkgoWriter.Printf("%s\n", outcmd)`. maybe better to use the same syntax everywhere.

And to be sure that get the logs even if an error occurs it could be better to use this:
```
outcmd, err := exec.Command("kubectl", "get", "bundles", "--all-namespaces").CombinedOutput()
GinkgoWriter.Printf("%s\n", outcmd)
Expect(err).To(Not(HaveOccurred()))
```

tests/e2e/logs_test.go
```diff
@@ -0,0 +1,47 @@
...
+			}
+			os.WriteFile("bundles-resource.log", outcmd, os.ModePerm)
+			outcmd, err = exec.Command("kubectl", "describe", "bundles", "--all-namespaces").CombinedOutput()
+			if err != nil {
+				fmt.Fprint(GinkgoWriter, err)
```
ldevulder, line 42: Same comment as above.

ldevulder (changes requested):

.github/workflows/master-e2e.yaml
juadk: yep exactly, actually I need to implement upstream logs to debug other issues in the CI.
I will add downstream logs later, I'm not able to quickly code scp go client etc...

tests/e2e/logs_test.go
juadk: Done!

ldevulder (changes requested):

tests/e2e/logs_test.go
```diff
@@ -0,0 +1,84 @@
...
+			}
+
+			_ = os.Mkdir("logs", 0755)
+			_ = os.Chdir("logs")
+			myDir, _ := os.Getwd()
+
+			var binaries []Binary = []Binary{elemental, logCollector}
+			for _, b := range binaries {
+				Eventually(func() error {
+					err := exec.Command("curl", "-L", b.Url, "-o", b.Name).Run()
```
ldevulder, line 51: And in fact  as `exec.Command` only returns an `error` you can simply write `return exec.Command("curl", "-L", b.Url, "-o", b.Name).Run()`.

tests/e2e/logs_test.go
```diff
@@ -0,0 +1,81 @@
...
+		By("Collecting additionals logs with kubectl commands", func() {
+			Bundles := getResourceLog{
+				"bundles",
+				[]string{"get", "describe"},
+			}
+
+			var getResources []getResourceLog = []getResourceLog{Bundles}
+			for _, r := range getResources {
+				for _, v := range r.Verb {
+					outcmd, err := exec.Command("kubectl", v, r.Name, "--all-namespaces").CombinedOutput()
```
ldevulder, line 73:
```suggestion
					outcmd, err := kubectl.Run(v, r.Name, "--all-namespaces")
```

tests/e2e/logs_test.go
```diff
@@ -0,0 +1,81 @@
+package e2e_test
+
+import (
+	"os"
+	"os/exec"
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"github.com/rancher/elemental/tests/e2e/helpers/misc"
```
ldevulder, line 10:
```suggestion
	"github.com/rancher-sandbox/ele-testhelpers/kubectl"
	"github.com/rancher/elemental/tests/e2e/helpers/misc"
```

ldevulder (approved):
Looks good!

I just forgot to point you to `kubectl.Run` function... I did some suggestions to add this.

Issue #577, opened 2022-12-13
Revisit elemental release procedures


davidcassany:
Few things to improve and elaborate around releases:

- [x] Not all sources are in github (specs, dockerfiles, etc.)
  * I'd suggest adding a some sort of `dist/obs/` folder within the repos to include those. In `dist/obs` there could be a README.md explaining the files in there are OBS specific and mostly used for SUSE's builds.
- [x] Rebuild of RPMs on PRs
- [x] Release candidate tags are useless, probably we should simply consider Stable as our RC once artifacts are in registry.suse.com
- [ ] #711
- [x] Arrange IBS project to built properly, SR is accepted, but images are not building
- [ ] #712

davidcassany:
I could not manage to test IBS as I wanted (mirroring registry.suse.com to registry.suse.de) but a clean deployment based on registry.suse.de artifacts went smooth.

I also struggled to set a custom certificate for registry.suse.de, looks like on every reboot the `/etc/rancher/k3s/registries.yaml` file is overwritten by something, looks like there is some k3s service or script creating this file on every reboot... 

Closing this issue as a couple of missing parts are separated in devoted non urgent or blocked issues.

PR #582, opened 2022-12-14
Add obs dockerfile and spec

davidcassany:
This PR adds the spec file and Dockerfile for `elemental` package and `node-image` image respectively. 

In addition it also introduces a new OBS workflow to build the package and image in a branch.

Part of #577


davidcassany:
Note these are added under the `.obs/dockerfile` and `.obs/specfile` subfolders. Finally I did not use a `dist/obs` folder because I realized there was already an `.obs` folder for the OBS SCM integration, so I thought of reusing this one just to include all OBS files in a single place.

Probably instead of `.obs/specfile` and `.obs/dockerfile` there could be something like `.obs/recipefiles`?

Itxaka (approved):

.obs/README.md
```diff
@@ -0,0 +1,8 @@
+# Open Build Service integration files
+
+This folder includes specific files to integrate with the Open Build Service (OBS) by
```
kkaempf, line 3:
```suggestion
This folder includes specific files to integrate with the [Open Build Service](https://build.opensuse.org) (OBS) by
```

.obs/README.md
```diff
@@ -0,0 +1,8 @@
...
+openSUSE. Includes the workflows definition file (`workflows.yml`) that defines
```
kkaempf, line 4:
```suggestion
[openSUSE](https://www.opensuse.org). Includes the workflows definition file (`workflows.yml`) that defines
```

.obs/dockerfile/Dockerfile
```diff
@@ -0,0 +1,61 @@
+# SPDX-License-Identifier: Apache-2.0
+# Define the names/tags of the container
+#!BuildTag: rancher/elemental-teal/5.3:latest
+#!BuildTag: rancher/elemental-teal/5.3:%VERSION%
+#!BuildTag: rancher/elemental-teal/5.3:%VERSION%-%RELEASE%
+#
+
+FROM suse/sle-micro-rancher/5.3:latest
```
kkaempf, line 8: I wonder if we shouldn't make `5.3` a build config macro 🤔

kkaempf (changes requested):

kkaempf (approved):

.obs/dockerfile/Dockerfile
davidcassany: Probably yes, but I would leave it for the time we need it. Hard to foresee now how it actually should be used.

PR #589, opened 2022-12-20
Deploy sample app through UI in Elemental cluster

### LINKED ISSUE #588: Make sure we can deploy an app in a fresh Elemental cluster
At the end of the UI e2e tests, we make sure that cluster's status is fine but it could even be better if we could try to deploy an app in the new cluster using the Rancher market place.

This new test will be a non regression test to not hit this bug again https://github.com/rancher/elemental-operator/issues/310

juadk:
Fix #588 
image

Verification run: :checkered_flag: 
https://github.com/rancher/elemental/actions/runs/3748671122

### DIFF
--- a/.github/workflows/master-e2e.yaml
+++ b/.github/workflows/master-e2e.yaml
@@ -236,6 +236,7 @@ jobs:
           SPEC: |
             cypress/e2e/unit_tests/machine_selector.spec.ts
             cypress/e2e/unit_tests/machine_inventory.spec.ts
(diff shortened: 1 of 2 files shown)


fgiudici (approved):
Looks good here!

juadk:
> Looks good here!

Actually, not ready yet, I need to change the code to being able switching namespace in the UI. It failed in the CI because by default, "Only User Namespace" is selected by default:
image

PR #595, opened 2022-12-28
Fix path for Cypress artifacts upload

juadk:
Cypress has been moved into the tests folder, so we also need to update the PATH for Cypress artifacts used in the upload action.

VR: https://github.com/rancher/elemental/actions/runs/3794824595


ldevulder (approved):
LGTM! Sorry for missing this...

ldevulder:
The failure of the VR is due to another issue and the videos are uploaded, so this PR is good and can be merged.

PR #598, opened 2022-12-30
ci: fix generated README for tests description

ldevulder:
We have to use a token to allow commit's push on main branch.

Tested on my lab after adding the same restrictions on my `main` branch:
- [test with branch locked and no token](https://github.com/ldevulder/gha-tests/actions/runs/3806093679)
- [test with branch locked and token used](https://github.com/ldevulder/gha-tests/actions/runs/3806100777/jobs/6474599340)

### DIFF
--- a/.github/workflows/update-tests-description.yaml
+++ b/.github/workflows/update-tests-description.yaml
@@ -12,6 +12,13 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v3
         with:
+          # A token is needed to be able to push on main, maybe this can be changed later
+          # with another GHA or with some webhook?
+          # This token is linked to a personal account (already used for GCP runner)
+          # So in case of token issue you have to check (no specific order and for example):
+          # - the expiration date
+          # - if the account associated still exists
+          # - if the person still has access to the repo
           token: ${{ secrets.PAT_TOKEN }}
       - name: Generate tests description file
         id: readme_generator
--- a/.github/workflows/update-tests-description.yaml
+++ b/.github/workflows/update-tests-description.yaml
@@ -11,6 +11,8 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v3
+        with:
+          token: ${{ secrets.PAT_TOKEN }}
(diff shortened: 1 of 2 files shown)


juadk (approved):

.github/workflows/update-tests-description.yaml
```diff
@@ -11,6 +11,8 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v3
+        with:
+          token: ${{ secrets.PAT_TOKEN }}
```
Itxaka, line 15: where is that Personal Access Token coming from? Cant we avoid that? rather use a different way of doing it as that its linked to the user that generated and then after a while we lost track of it...

ldevulder: Yes but I found no other ways, and moreover we already have a token (the same) used for GCP self-hosted runner. For the runner it might be possible to use webhooks but this will need more time consuming tests. We can check to remove it later I think. Anyway if the tests description update doesn't work anymore this is not something that will block us.

Itxaka: Cool for me, I already mentioned it. If it breaks I wont be here to fix it so I dont care.... joking of course :)

Maybe adding that info to the readme.md so when it breaks we know how to proceed?

Itxaka (approved):

.github/workflows/update-tests-description.yaml
ldevulder: More useful in the workflow file for the info I think, no? Not sure that someone will check in `tests/README.md` file in case of issue in the GH action.

Tell me what do you think it's better, I will then add a note and push a commit for this before merging.

Itxaka: either way, as long as its written somewhere its good for me :=)

ldevulder: Added with this [commit](https://github.com/rancher/elemental/pull/598/commits/aacc78f3014ae801ec4c9ea50837086a8a6b9964).

PR #607, opened 2023-01-05
Update gitignore file

juadk:
Cypress folder was moved from root to tests directory and we forgot to update the `.gitignore` file.


fgiudici (approved):
sure!

PR #608, opened 2023-01-06
e2e: add parallel VM installation

### LINKED ISSUE #148: e2e: try to parallelize VM installation in Ginkgo e2 tests
This could be good to either speed up the installation/test and a good way to test parallel node installation in Rancher Manager.

See this Ginkgo doc: https://onsi.github.io/ginkgo/#spec-parallelization

ldevulder:
Should fix #148.

Verification run:
- [OBS-Dev-K3s-E2E](https://github.com/rancher/elemental/actions/runs/3956988868)
- [OBS-Dev-K3s-E2E with 10 nodes](https://github.com/rancher/elemental/actions/runs/3956200426) - seems to fail, but I don't know why exactly
- [OBS-Dev-RKE2-E2E](https://github.com/rancher/elemental/actions/runs/3950719315)
- [OBS-Dev-RKE2-E2E with 10 nodes](https://github.com/rancher/elemental/actions/runs/3952511906)

**NOTE:** I also tested with 20 nodes but it fails all the time and also with 10 nodes on K3s. But as the CI version (with 5 nodes) works well I think that we can merge this PR and I will check why more nodes are failing and fix this in another PR.


.github/workflows/master-e2e.yaml
```diff
@@ -316,23 +311,19 @@ jobs:
...
           VM_INDEX: 2
         run: cd tests && make e2e-upgrade-node
-      - name: Upgrade node 3 (with managedOSVersionName method) to specified Teal version
-        if: inputs.test_type == 'cli'
-        env:
-          IMAGE_VERSION: teal-5.3
-          UPGRADE_TYPE: managedOSVersionName
-          VM_INDEX: 3
-        run: cd tests && make e2e-upgrade-node
+#     - name: Upgrade node 3 (with managedOSVersionName method) to specified Teal version
```
frelon, line 314: Maybe we can remove these commented lines?

.github/workflows/master-e2e.yaml
```diff
@@ -297,17 +297,12 @@ jobs:
           UPGRADE_TYPE: osImage
           VM_INDEX: 1
         run: cd tests && make e2e-upgrade-node
-      - name: Bootstrap node 2 with current build (use ISO)
+      - name: Bootstrap node 2 and 3 with current build (use ISO)
         if: inputs.test_type == 'cli'
         env:
           ISO_BOOT: true
           VM_INDEX: 2
```
frelon, line 304: Is the `VM_INDEX: 2` still valid?

davidcassany (approved):
Looks good to me :+1:

.github/workflows/master-e2e.yaml
ldevulder: The idea is to re-add this test shortly, but yes I can remove them, we still have the history.

.github/workflows/master-e2e.yaml
ldevulder: Yes, `VM_INDEX` is the number where start (so node 2) and `VM_NUMBERS` is the total number of node to have at the end, so 3 here. ANd in fact maybe better to rename the variables to `NODE_*` instead of `VM_*`, if we have the test running on *real* HW later. But this can be done later too.

.github/workflows/master-e2e.yaml
ldevulder: done!

frelon (approved):

PR #610, opened 2023-01-09
Install kernel-firmware-all package

kkaempf:

### DIFF
--- a/.obs/dockerfile/Dockerfile
+++ b/.obs/dockerfile/Dockerfile
@@ -11,7 +11,7 @@ FROM suse/sle-micro-rancher/5.3:latest
 # kernel-firmware-amdgpu installed here until 5.4 is released and includes it
 RUN zypper in -y -- \
   NetworkManager-wwan \
-  kernel-firmware-amdgpu \
+  kernel-firmware-all \
   -systemd-presets-branding-SLE-Micro-for-Rancher \
   elemental
 


Itxaka:
This affects the image size by a lot, doesn't it? Maybe we should have a slim and full images? Otherwise we are talking about applying updates that are what, 900mb? A bit big

Itxaka (approved):

kkaempf:
> This affects the image size by a lot, doesn't it? Maybe we should have a slim and full images? Otherwise we are talking about applying updates that are what, 900mb? A bit big

Agreed - but what is 'slim' and what is 'big' ? There's always something missing and something superfluous 🤷🏻 
/cc @agracey

agracey:
> but what is 'slim' and what is 'big'

My naive suggestion would be:
- Slim is the most minimal that can boot on common hardware (VMs, NUCs, raspberry pis, etc)
- Full is everything (opensource) you would expect to need in 90% of environments 


Then, most people would start with full to learn how the tooling works then move to the slim and add additional packages.

kkaempf:
> Then, most people would start with full to learn how the tooling works then move to the slim and add additional packages.

I fear we need a way to support such setups 🙈

kkaempf:
> Would it make sense to schedule a sprint after the next release looking at image size and varietals?

You can see image sizes [here](https://download.opensuse.org/repositories/isv:/Rancher:/Elemental:/Dev:/Teal53/media/iso/) (iso) and [here](https://download.opensuse.org/repositories/isv:/Rancher:/Elemental:/Dev:/Teal53/images/) (Raspberry image).

If image sizes are a concern, we can add a card to look at minimizing the image (by removing unneeded stuff pulled in by rpm deps)

PR #611, opened 2023-01-11
Add support for Podman for registration injection into ISO

rdoxenham:
Here I've added support for using Podman for the injection of registration information to the Elemental Teal ISO. This does not replace Docker as the default (if it's available on the client) but provides an option for those of us using Podman instead (bypassing the `alias docker=podman` hack). This has been tested on Tumbleweed with Podman and Ubuntu with Docker.

I have also cleaned up a little bit of whitespace that I found lingering.

### DIFF
--- a/.github/elemental-iso-add-registration
+++ b/.github/elemental-iso-add-registration
@@ -47,7 +47,7 @@ iso_repack() {
 
   cp "${REG_CONFIG_FILE}" "${TEMPDIR}/config/${REG_FILE}"
 
-  if [ -f "/var/run/docker.pid" ]; then
+  if [ -S "/var/run/docker.sock" ]; then
     CONT_ENGINE="docker"
   elif [ -x "$(command -v podman)" ]; then
     CONT_ENGINE="podman"
--- a/.github/elemental-iso-add-registration
+++ b/.github/elemental-iso-add-registration
@@ -6,7 +6,7 @@ usage() {
 me=$(basename "$0")
 cat << EOF
 
-usage: 
+usage:
    $me REGISTRATION_CONFIG_FILE [ISO_URL]
 
 REGISTRATION_CONFIG_FILE is the yaml file required by elemental-register in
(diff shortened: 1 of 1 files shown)


davidcassany (approved):
Thanks for the contribution :+1: Looks good to me

Itxaka (approved):

fgiudici:
Hey @rdoxenham , thanks for the PR!
unfortunately the test on MAC is failing: seems the socket check for docker on MAC fails.

I cannot thing about something compact like the -S though 🤔 

What about testing for the binary instead? That should be more robust. Something like:

```
is_installed() {
        cmd="$1"
        $cmd 2>&1 > /dev/null
        return $?
}

if is_installed "docker -v"; then
...
```

Any better idea more than welcome 😄

davidcassany:
IMHO we should just merge it, the macOS issue looks like an unrelated issue.

rdoxenham:
> Hey @rdoxenham , thanks for the PR! unfortunately the test on MAC is failing: seems the socket check for docker on MAC fails.
> 
> I cannot thing about something compact like the -S though thinking
> 
> What about testing for the binary instead? That should be more robust. Something like:
> 
> ```
> is_installed() {
>         cmd="$1"
>         $cmd 2>&1 > /dev/null
>         return $?
> }
> 
> if is_installed "docker -v"; then
> ...
> ```
> 
> Any better idea more than welcome smile

Yeah, I think this is a problem with the way that the MacOS test is being executed. On MacOS with Docker there is the socket available in the same place as Linux, so this is checked properly:

```
rdo@macbook ~ % file /var/run/docker.sock
/var/run/docker.sock: socket
```

I don't think it's clean enough to check that the docker package is installed, even though my own commit just checks if `podman` is available and executable with `elif [ -x "$(command -v podman)" ]; then` when it fails to find the Docker socket.

rdoxenham:
Okay, after further investigation this is because the GitHub tests use https://github.com/abiosoft/colima which doesn't create the Docker socket at /var/run/docker.sock. There's also an issue (https://github.com/abiosoft/colima/issues/503) with volume mounts with Colima which many people are complaining about, so the tests may still fail, depending on the Colima configuration.

fgiudici (approved):
Many thanks @rdoxenham !

PR #612, opened 2023-01-11
ci: fix "test inject in iso" workflow

ldevulder:
It fails on macOS, as Docker needs to be installed before buildx.

### DIFF
--- a/.github/elemental-iso-add-registration
+++ b/.github/elemental-iso-add-registration
@@ -25,7 +25,7 @@ abort() {
 }
 
 iso_repack() {
-  local TEMPDIR=$(mktemp -d)
+  local TEMPDIR=$(mktemp -d $HOME/elemental_iso.XXX)
   local ISO_FILE=$(basename "${ISO_URL}")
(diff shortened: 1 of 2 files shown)


.github/elemental-iso-add-registration
```diff
@@ -25,7 +25,7 @@ abort() {
 }
 
 iso_repack() {
-  local TEMPDIR=$(mktemp -d)
+  local TEMPDIR=$(mktemp -d $HOME/elemental_iso.XXX)
```
fgiudici, line 28: just wondering: why do we need this change? 🤔

fgiudici (approved):
Looks good here!

.github/elemental-iso-add-registration
ldevulder: Mecause of MacOS, @juadk did some test and on mac it tries to create the temp directory in `/var/folders`... Don't know why... But this trick works both on Linux and Mac.
Maybe we should have add a comment here yes...

PR #613, opened 2023-01-16
Fix ISO build and upgrade tests with managedOSVersionChannel 

ldevulder:
`teal-5.2` image doesn't exist anymore, better to use a `fake-image` for UI upgrade test. This PR also fix the ISO build by using the correct images.


davidcassany (approved):
LGTM

PR #617, opened 2023-01-17
ci/integ: fix failing test

ldevulder:
The checked text has been slightly modified due to refactoring in elemental-cli. See https://github.com/rancher/elemental-cli/pull/404.


davidcassany (approved):
Thanks for the fix! I did not realize it could have impact that area...

PR #630, opened 2023-01-24
ci/e2e: don't run test on latest RM on every push

ldevulder:
This worflow is scheduled *after* build-ci to be sure that we have the lastest version built. The scheduling is also to avoid running the worklow on each push on main.


juadk (approved):
LGTM thanks

PR #642, opened 2023-02-02
add-registration: clean up upon abort

fgiudici:
If we abort, we will leave around the temp directory.
Moreover, there is an ISO in that directory that could be mistakenly taken as the ISO with the combined registration.


kkaempf (approved):
:+1:

PR #644, opened 2023-02-06
ci/e2e: increase timeout when delete resource in UI

### LINKED ISSUE #643: e2e UI - We do not wait enough after removing machine registration
Sometime, CI fails because the machine registration we delete it still print on the screen 10 seconds after.

https://github.com/rancher/elemental/actions/runs/4100331557

Machine registration testing -- Delete machine registration (failed)

We should increase timeout to avoid flaky issue.

juadk:
Fix #643 

Verification run:
https://github.com/rancher/elemental/actions/runs/4101679988 :heavy_check_mark: 

Wait 20 sec instead of 10 sec when we delete a machine registration in our e2e UI tests.
I also remove a useless 8 seconds timeout, as the default is 10, we do not need a lower timeout.


fgiudici (approved):
looks good here!

PR #652, opened 2023-02-07
Add shutdown-k3s.service to OBS spec

frelon:
Adding the service to %pre, %post, %preun and %postun


davidcassany (approved):
LGTM

PR #681, opened 2023-02-17
ci/ui: add label functionality to Cypress tests

### LINKED ISSUE #680: e2e/ui: Add labels to filter Cypress tests
As we will split upgrade tests from our main scenario to a dedicated test, we need a way to tell to Cypress which tests we want to load. Mainly because we do not want to do all the tests in the upgrade scenario, we want only update there.

I tried to define tests as worklfow variables but it to borring to maintain.. we will have to think about updating every workflow files... 
image

There is a smarter solution using Labels but it's not delivered  in Cypress code.
Method 2 [here](https://testersdock.com/cypress-test-tags/)

juadk:
Fix #680 

Verification runs:
[K3s-UI_E2E-Stable_RM](https://github.com/rancher/elemental/actions/runs/4225773365) :heavy_check_mark: 
[K3s-UI_E2E-Latest_RM](https://github.com/rancher/elemental/actions/runs/4224597044) :heavy_check_mark:

### DIFF
--- a/tests/cypress/e2e/unit_tests/advanced_filtering.spec.ts
+++ b/tests/cypress/e2e/unit_tests/advanced_filtering.spec.ts
@@ -16,53 +16,56 @@ import { TopLevelMenu } from '~/cypress/support/toplevelmenu';
 import '~/cypress/support/functions';
 import { Elemental } from '~/cypress/support/elemental';
 import { contains } from 'cypress/types/jquery';
+import filterTests from '~/cypress/support/filterTests.js';
 
-Cypress.config();
-describe('Advanced filtering testing', () => {
-  const topLevelMenu   = new TopLevelMenu();
-  const elemental      = new Elemental();
-  const ui_account     = Cypress.env('ui_account');
-  const elemental_user = "elemental-user"
-  const ui_password    = "rancherpassword"
-
-  beforeEach(() => {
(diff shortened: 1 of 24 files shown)


tests/cypress/e2e/unit_tests/advanced_filtering.spec.ts
```diff
@@ -16,53 +16,56 @@ import { TopLevelMenu } from '~/cypress/support/toplevelmenu';
 import '~/cypress/support/functions';
 import { Elemental } from '~/cypress/support/elemental';
 import { contains } from 'cypress/types/jquery';
+import filterTests from '~/cypress/support/filterTests.js';
```
ldevulder, line 19: Maybe more coherent to do the import like this (without the `.js`), no?
```javascript
import filterTests from '~/cypress/support/filterTests';
```

ldevulder (approved):
LGTM, lot of changes due to the tags, but only one comment (but it applies for all `import` ;-)).

tests/cypress/e2e/unit_tests/advanced_filtering.spec.ts
juadk: I tried it one more time and looks like `.js` must be used because it is imported inside a `.ts` file.

PR #692, opened 2023-02-23
No concurrent suc upgrade

davidcassany:
This partially fixes rancher/elemental-operator#364. It prevent two concurrent upgrade procedures within the same node. Since this is at node level this is not a full fix as two different upgrade procedures happening at the same time in different nodes of the same cluster is still possible. IMHO the solutions should focus of preventing any concurrency at cluster level.

In any case this solves the ugliest issue, which is having two simultaneous `elemental upgrade` calls over the same node simultaneously.

### DIFF
--- a/framework/files/usr/sbin/suc-upgrade
+++ b/framework/files/usr/sbin/suc-upgrade
@@ -11,14 +11,12 @@ function config()
     if [ ! -s $CONF_FILE ]; then
         if [ -e ${HOST_DIR}/oem/90_operator.yaml ]; then
             rm -f ${HOST_DIR}/oem/90_operator.yaml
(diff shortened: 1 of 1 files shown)


ldevulder (approved):
I can't test manually but the code looks good!

davidcassany:
> I can't test manually but the code looks good!

Yes I know, I did some manual tests (took me a while to prepare everything) I think the best will be merging and then testing in Dev, if succeeds then tagging to staging and we will have everything for a proper testing.

frelon (approved):
Haven't tested it but looks better than the older version :+1:

fgiudici (approved):
let's merge!

PR #699, opened 2023-02-28
Do not allow downgrades without force flag

davidcassany:
This commit sorts the current OS version with the one we aim to upgrade to. If the OS to be upgraded to is not higher than the current one it assumes we are up to date and the upgrade script exits 0. This causes the upgrade plan to succeed and to be considered as upgraded. This is to fix https://github.com/rancher/elemental-operator/issues/364#issuecomment-1443139017

Alternatively if a downgrade is required it is still possible, but requires a customized `Update Group` (`ManagedOSImage` crd) to include the `FORCE` environment variable within the upgrade script:

```yaml
apiVersion: elemental.cattle.io/v1beta1
kind: ManagedOSImage
metadata:
  name: upgrade3
  namespace: fleet-default
spec:
  clusterTargets:
    - clusterName: my-cluster
  upgradeContainer:
    command:
      - "/usr/sbin/suc-upgrade"
    envs:
      - name: FORCE
        value: "true"
    image: myimage:mytag
```

With the `FORCE` flag no version is taken into account and the image is applied in any case.

`elemental-operator` logic makes use of the `osImage` field if no `upgradeContainer` field is provided to actually create a default `ugpradeContainer`. However if `upgradeContainer` is provided then `osImage` field is ignored. I'd say this logic is not very intuitive, but I'd say this should be handled separately if we consider there is something to change in there.

Fixes rancher/elemental-operator#364

### DIFF
--- a/framework/files/usr/sbin/suc-upgrade
+++ b/framework/files/usr/sbin/suc-upgrade
@@ -20,6 +20,24 @@ function config()
     fi
 }
 
+function isHigherVersion()
+{
+    local img_ver=$( . ${RELEASE_FILE} && printf '%s\n' "${IMAGE_TAG}" )
+    local host_ver=$( . ${HOST_DIR}${RELEASE_FILE} && printf '%s\n' "${IMAGE_TAG}" )
+    local higher_ver
+
+    # Without knowing the version in the host, all img versions are
+    # considered higher
+    [ -z "${host_ver}" ] && return 0
+    
+    [ "${host_ver}" == "${img_ver}" ] && return 1
+
+    # Note sort -V is a natural numbering sort, not semver
+    higher_ver=$(printf '%s\n' "${img_ver}" "${host_ver}" | sort -rV | head -n1)
+    [ "${higher_ver}" == "${img_ver}" ] && return 0
+    return 1
+}
+
 (
   flock -w $LOCK_TIMEOUT 200 || exit 1
   if ! nsenter -i -m -t 1 -- systemctl is-system-running; then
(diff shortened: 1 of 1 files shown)


davidcassany:
@kkaempf @agracey this is the fix I tested to skip downgrades by default. I'd be also fine by not including it and assume the admin has to provide sane update groups and/or try to fix the issue beyond elemental (system-upgrade-controller, fleet, rancher, etc.). IMHO the good solution would be preventing concurrent upgrades at downstream cluster level, this is essentially orchestrated by the fleet and system-upgrade-controller pair.

agracey:
I think this makes sense for now. Since there is a workaround to force a downgrade (however unpleasant), I'm comfortable with the limitation

davidcassany:
> I think this makes sense for now. Since there is a workaround to force a downgrade (however unpleasant), I'm comfortable with the limitation

Yes, we can improve it by making the `ManagedOSImage` logic a bit more _intuitive_. So we could enable something like:

Current common `Upgrade Group` without force:
```yaml
apiVersion: elemental.cattle.io/v1beta1
kind: ManagedOSImage
metadata:
  name: upgrade3
  namespace: fleet-default
spec:
  clusterTargets:
    - clusterName: my-cluster
  osImage: myimage:mytag
```

`Upgrade Group` with force:
```yaml
apiVersion: elemental.cattle.io/v1beta1
kind: ManagedOSImage
metadata:
  name: upgrade3
  namespace: fleet-default
spec:
  clusterTargets:
    - clusterName: my-cluster
  osImage: myimage:mytag
  upgradeContainer:
    envs:
    - name: FORCE
      value: "true"
```

So basically use `osImage` as `upgradeContainer.image` even if some other fields of `upgradeContainer` are provided. Alternatively we could also add a new `forceUpgrade` field at the level of `osImage`, but probably I'd wait for that until we have further or more refined feedback on what is required for the upgrades. If we add a new field we will have to support it and I dislike having multiple ways of setting the exact same thing.

fgiudici (approved):
looks good here

Issue #667, opened 2023-02-14
k3s/rke2 is not deployed on the node (sporadic)


ldevulder:
**What steps did you take and what happened:**
A node is successfully provisioned with Elemental but after the reboot `element-system-agent` doesn't connect to Rancher Manager and the configuration/cluster deployment is not done. For example the hostname is not set as well as the configured user/password.

It's something very sporadic that I have sometimes during manual testing and it also seems to happen from time to time on the CI ([here](https://github.com/rancher/elemental/actions/runs/4173563257/jobs/7228158013) for example).

**What did you expect to happen:**
The node to be added in the cluster.

**Anything else you would like to add:**
Find attached some logs from elemental-support and journalctl.
[journal.log](https://github.com/rancher/elemental/files/10733987/journal.log)
[rancher-8050-2023-02-14T150600Z.tar.gz](https://github.com/rancher/elemental/files/10733988/rancher-8050-2023-02-14T150600Z.tar.gz)

**Environment:**

- Elemental release version (use `cat /etc/os-release`): Dev version
- Rancher version: 2.7.1
- Kubernetes version (use `kubectl version`): N/A
- Cloud provider or hardware configuration: VM on KVM

ldevulder:
I had the issue again on a 15 nodes cluster in the CI: https://github.com/rancher/elemental/actions/runs/4312380600/jobs/7523570584.
Some logs are available here: https://github.com/rancher/elemental/actions/runs/4312380600

Here also logs taken with `elemental-support`:
- from the problematic node: 
[node-63c74412-8c70-47e4-a98f-3e47c68de602-2023-03-02T103611Z.tar.gz](https://github.com/rancher/elemental/files/10870449/node-63c74412-8c70-47e4-a98f-3e47c68de602-2023-03-02T103611Z.tar.gz)

- from the 1st master node:
[node-38892689-eb92-4237-bc33-132ac2daa0c0-2023-03-02T103850Z.tar.gz](https://github.com/rancher/elemental/files/10870450/node-38892689-eb92-4237-bc33-132ac2daa0c0-2023-03-02T103850Z.tar.gz)

kkaempf:
adding @moio

kkaempf:
What are the hardware specs of the nodes, esp. the (speed of) storage used ?
Are you trying to deploy K3S or RKE2 ?

kkaempf:
> Are you trying to deploy K3S or RKE2 ?

From the logs, I looks like RKE2

[...]

PR #718, opened 2023-03-07
ci/e2e: multiple changes to improve scalability tests

ldevulder:
Multiple changes to improve scalability tests:
- Use LVM with striping I/O for local storage, this will improve I/O performance for bootstrapping nodes when multiples local SSDs are used (which is the case for highmem runner).
- Bootstrap additional nodes with ISO and use PXE to bootstrap the master ones.
- Wait a little bit before rebooting the nodes, this is to avoid starting all VMs at the same time.

Verification runs:
- [OBS-Stable-RKE2-E2E with 30 nodes](https://github.com/rancher/elemental/actions/runs/4365471173)
- [OBS-Stable-RKE2-E2E with 50 nodes](https://github.com/rancher/elemental/actions/runs/4375907053)

**NOTE:** VR with 50 nodes is failing because of issue #667, not related to this PR.

### DIFF
--- a/.github/workflows/master-e2e.yaml
+++ b/.github/workflows/master-e2e.yaml
@@ -323,11 +323,10 @@ jobs:
       - name: Configure Rancher & Libvirt
         if: inputs.test_type == 'cli'
         run: cd tests && make e2e-configure-rancher
-      - name: Bootstrap node 1, 2 and 3 (use and ISO) in pool "master" (use Emulated TPM if possible)
+      - name: Bootstrap node 1, 2 and 3 (with iPXE) in pool "master" (use Emulated TPM if possible)
         if: inputs.test_type == 'cli'
         env:
           EMULATE_TPM: true
-          ISO_BOOT: true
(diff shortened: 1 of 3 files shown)


.github/workflows/master-e2e.yaml
```diff
@@ -369,6 +368,7 @@ jobs:
       - name: Bootstrap additional nodes (use iPXE) in pool "worker" (total of ${{ inputs.node_number }})
```
juadk, line 368: I think this comment has to be updated as nodes in worker pool are now installed via ISO.
```suggestion
      - name: Bootstrap additional nodes (use ISO) in pool "worker" (total of ${{ inputs.node_number }})
```

juadk (approved):

.github/workflows/master-e2e.yaml
ldevulder: Done!

juadk (approved):

PR #723, opened 2023-03-10
Run elemental-register in systemd service

### LINKED ISSUE #722: Elemental-register is called without inheriting `/etc/profile` variables and settings

It has been noted that `/etc/sysconfig/proxy` is not taken into accound when calling `elemental-register` from the cloud-config setup as part of the `network.after` yip stage.

The proxy been taken into account for the `elemental-register` call, otherwise the registration just fails and the process is stuck trying to register in the foreground of `cos-setup-network.service`.

This is happening because `/etc/sysconfig/proxy` is loaded as part of the `/etc/profile` which is not being loaded on the shell executing the commands defined in  `network.after` stage.

Within the `network.after` I did few trials with the following commands:

```bash
env
systemd-cat -t elemental_test env
bash -cl env
```

Only the last one is including the variables and setup defined in `/etc/profile`.

Within this experiments I also noted `elemental-register` is executed in the foreground and it does not end until it registers (keeps trying on failures). This is bad, as this means the `network.after` stage is stuck with this command and anything after it is not executed. This also means `cos-setup-network` service is stuck in activating state. I realized that because the testing commands I added in `network.after` in `iso-config` where not executed as before those the same stage from `/system/oem` is executed and got stuck there. We should be careful to not add blocking commands in yip stages.

I believe the solution is to mimic the same approach as with elemental-system-agent. Wrap elemental-register in a unit file (called with  `bash --login -c elemental-register ...`) and simply start the service as part of the cloud-init setup. Also we could consider calling elemental-register on every boot and just enable the service by default, but I'd do that on a second interation as this might require some small tweaks in elemental-register code to handle this case appropriately.

**Environment:**

- Elemental release version (use `cat /etc/os-release`): Staging
- Rancher version:
- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:

davidcassany:
Fixes #722

### DIFF
--- a/framework/files/usr/lib/systemd/system/elemental-register.service
+++ b/framework/files/usr/lib/systemd/system/elemental-register.service
@@ -6,5 +6,5 @@ After=network-online.target
 
 [Service]
 Type=oneshot
-ExecStart=bash --login -c elemental-register --debug /run/initramfs/live
+ExecStart=bash --login -c "elemental-register --debug /run/initramfs/live"
 
--- a/framework/files/usr/lib/systemd/system/elemental-register.service
+++ b/framework/files/usr/lib/systemd/system/elemental-register.service
@@ -1,0 +1,10 @@
+[Unit]
+Description=Elemental Register
+Documentation=https://elemental.docs.rancher.com
+Wants=network-online.target
+After=network-online.target
+
+[Service]
+Type=oneshot
+ExecStart=bash --login -c elemental-register --debug /run/initramfs/live
+
--- a/.obs/specfile/elemental.spec
+++ b/.obs/specfile/elemental.spec
@@ -77,18 +77,22 @@ rm -rf %{buildroot}/usr/libexec/.placeholder
 %pre
 %service_add_pre elemental-populate-node-labels.service
 %service_add_pre shutdown-containerd.service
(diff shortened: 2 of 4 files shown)


framework/files/usr/lib/systemd/system/elemental-register.service
```diff
@@ -0,0 +1,10 @@
+[Unit]
+Description=Elemental Register
+Documentation=https://elemental.docs.rancher.com
+Wants=network-online.target
+After=network-onlune.target
```
frelon, line 5: Typo: onlune

frelon (commented):
Small typo and should probably update the motd for watching the logs

framework/files/usr/lib/systemd/system/elemental-register.service
davidcassany: 🤦🏽

davidcassany:
I still have not test it, as requires a build of many pieces. But I do not expect major surprises, gonna try to test it today.

davidcassany:
I tested it manually and it worked fine. IMHO this is ready to be reviewed and eventually merged.

frelon (approved):
LGTM

PR #724, opened 2023-03-10
ci/e2e: add a sequential option to boostrap nodes

ldevulder:
We have to cover most cases: parallel bootstrapping as well as sequential ones, as we can have different behaviors and so different issues.


.github/workflows/cli-obs-manual-workflow.yaml
```diff
@@ -32,6 +32,10 @@ on:
         description: Runner template to use
         default: elemental-e2e-ci-runner-spot-x86-64-template-n2-standard-16-v2
         type: string
+      sequential:
+        description: Defines if bootstrapping is done sequantially (true) or in parallel (false)
```
juadk, line 36:
```suggestion
        description: Defines if bootstrapping is done sequentially (true) or in parallel (false)
```

.github/workflows/master-e2e.yaml
```diff
@@ -84,6 +84,10 @@ on:
         description: Runner template to use
         default: elemental-e2e-ci-runner-spot-x86-64-template-n2-standard-16-v2
         type: string
+      sequential:
+        description: Defines if bootstrapping is done sequantially (true) or in parallel (false)
```
juadk, line 88:
```suggestion
        description: Defines if bootstrapping is done sequentially (true) or in parallel (false)
```

juadk (approved):
LGTM, just a small typo.

juadk (approved):

PR #725, opened 2023-03-10
ci: add worflow to manually trigger a test

ldevulder:
This allows anyone to manually trigger a test on any version of Elemental ISO/operator/kubernetes client cluster.

Default is set to OBS-Dev on K3s.

### DIFF
--- a/.github/workflows/cli-obs-manual-workflow.yaml
+++ b/.github/workflows/cli-obs-manual-workflow.yaml
@@ -1,0 +1,60 @@
+# This workflow calls the master E2E workflow with custom variables
+name: CLI-OBS-Manual-Workflow
+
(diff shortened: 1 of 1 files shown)


davidcassany (approved):
Great! Thanks much!

So to use some other project we just have to provide the URLs  and values for the defined inputs right?

ldevulder:
> So to use some other project we just have to provide the URLs and values for the defined inputs right?

After merging this PR you should be able to trigger the `CLI-OBS-Manual-Workflow` workflow on any repository by setting `iso_to_test` and `upgrade_operator` to the version you want. (Yes `upgrade_operator` is not a so good name, this will changed in the future).

Issue #732, opened 2023-03-13
Persistent paths not mounted at boot


davidcassany:
**What steps did you take and what happened:**

The issue raised on a 50 nodes deployment on the CI. There was one node that did not join the cluster. This happened because the immutable-rootfs layout failed to properly mount the persistent partition and all of the state RW paths depending on it. Once booted the `bootstrap.sh` script failed to install RKE2 as /opt was not writable.

**What did you expect to happen:**
The node boostraps RKE2 and joins the cluster. Probably we should just fail at boot time, immutable-rootfs simply omitted to create the persistent mountpoints and continued to boot normally, probably better to include all them in /etc/fstab and let systemd fail on later stages if the underlaying devices and mountpoints do not appear.

Worth to mention that on boot `/usr/local` got mounted based on `/etc/fstab` contents after switching root when systemd took control of the new root `/`.

**Anything else you would like to add:**

Journalctl logs of the initrd boot stage contain most of the relevant info:

```
Mar 13 09:14:12 localhost elemental[784]: INFO[2023-03-13T09:14:12Z] Executing /system/oem/01_elemental-rootfs.yaml
Mar 13 09:14:12 localhost elemental[784]: INFO[2023-03-13T09:14:12Z] Applying 'Elemental Rootfs Layout Settings' for stage 'rootfs.after'. Total stages: 1
Mar 13 09:14:12 localhost elemental[784]: INFO[2023-03-13T09:14:12Z] Processing stage step 'Grow persistent'. ( commands: 0, files: 0, ... )
Mar 13 09:14:12 localhost elemental[784]: INFO[2023-03-13T09:14:12Z] Extending last partition up to 0 MiB
Mar 13 09:14:13 localhost systemd-udevd[601]: sda5: Failed to add device '/dev/sda5' to watch: Operation not permitted
Mar 13 09:14:13 localhost elemental[784]: INFO[2023-03-13T09:14:13Z] Stage 'rootfs.after'. Defined stages: 1. Errors: false
Mar 13 09:14:13 localhost elemental[784]: INFO[2023-03-13T09:14:13Z] Executing /system/oem/04_accounting.yaml
Mar 13 09:14:13 localhost elemental[784]: INFO[2023-03-13T09:14:13Z] Executing /system/oem/05_motd_and_autologin.yaml
Mar 13 09:14:13 localhost elemental[784]: INFO[2023-03-13T09:14:13Z] Executing /system/oem/05_network.yaml
[...]

PR #740, opened 2023-03-16
Do not expand COS_PERSISTENT by default

davidcassany:
This commit drops the expansion cloud-init step for COS_PERSISTENT partition. The installation already uses all available size by default there is not need to attempt a resize on every boot.

In addition this commit also updates the persistent volumes configuration to be aligned with the latest elemental-toolkit changes. Now volumes are set as part of the kernel arguments at boot and not as part of the cloud-init setup. This prevents including filesystem label literals within the immutable OS.

Related to #732


frelon (approved):
Very nice!

ldevulder (approved):

davidcassany:
I blocked it as I'd like to merge it after verifying rancher/elemental-toolkit#1743 solves the issue. The patch here I believe it prevents the race condition to happen so we would not hit the corner case that the mention PR aims to fix.

PR #752, opened 2023-03-21
Include cryptsetup in elemental-teal

### LINKED ISSUE #749: Include utilities for encrypted volumes and enable longhorn encrypted storage
Enable encrypted longhorn storage. For this luks utilities such as cryptsetup should be included and validated.

Elemental-teal image does not include luks, hence IMHO this should be included by default as managing encrypted volumes should be possible with elemental-teal without having to relay on the inclusion of additional software at installation time.

**Environment:**

Luks is missing for all versions.

- Elemental release version (use `cat /etc/os-release`): 
- Rancher version:
- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:

davidcassany:
Fixes #749


frelon (approved):
Very nice! 👍

PR #768, opened 2023-03-30
Load proxy setup from sysconfig and allow elemental specific env variables

davidcassany:


davidcassany (commented):
I am not 100% sure about reading `/etc/sysconfig/proxy` for two reasons, it is SUSE specific thing and it is use case specific. 

On the other side this makes the SUSE proxy setup transparent to elemental. Then the downstream clusters would require rancher specific setup which can be easily done from the UI as part of the cluster creation. I think this makes the proxy story pretty smooth.

juadk:
> I am not 100% sure about reading `/etc/sysconfig/proxy` for two reasons, it is SUSE specific thing and it is use case specific.
> 
> On the other side this makes the SUSE proxy setup transparent to elemental. Then the downstream clusters would require rancher specific setup which can be easily done from the UI as part of the cluster creation. I think this makes the proxy story pretty smooth.

Otherwise we could have put proxy settings in `/etc/sysconfig/elemental` but it does not make a lot of sense because that means user will have to define the proxy elsewhere if he needs it...
I'm in favor of using `/etc/sysconfig/proxy` in a first time and check later if someone comes with a different base image.

davidcassany:
A testing ISO including the changes can be found here 
```
https://download.opensuse.org/repositories/home:/dcassany:/Elemental:/Dev:/Teal53/media/iso/elemental-teal.x86_64.iso
``` 

And the helm chart at:

```
oci://registry.opensuse.org/home/dcassany/elemental/dev/charts/rancher/elemental-operator-chart
```

davidcassany:
For some reason the static link points to an old ISO... 🤔 That makes me wonder if ISO publishing is really functional. I think @kkaempf saw something similar a while ago, probably this is caused by some cache around 🤷🏽‍♂️ If I download from the link including the build number it just works fine and the ISO is at a different version, this was not supposed to happen, gonna check tomorrow morning if this is still happen, if so we should investigate it, otherwise probably this is just a transitory OBS glitch.

juadk:
All good this morning!
`elemental-register` went through the proxy and this is the `elemental-system-agent` file:
image

Nothing in `/etc/sysconfig/proxy` but that was expected because we only configured it for `seed-image` and I think we do not need it because Rancher will be able to reach the host and send the config with proxy settings.

Last thing, I was able to create a cluster.

juadk (approved):

davidcassany:
> All good this morning!

Yes, I'd say we were hitting some cache somewhere... 🤷🏽‍♂️ Probably because the filename does not change we are causing some files update delay as there is no 404 on HEAD requests...

> Nothing in /etc/sysconfig/proxy but that was expected because we only configured it for seed-image and I think we do not need it because Rancher will be able to reach the host and send the config with proxy settings.

I am not sure I follow that... anyway, my understanding of the situation is that setting the  `/etc/sysconfig/proxy` is required within the live system to make sure `elemental-register` is capable to reach rancher manager. Also the `/etc/sysconfig/proxy` I believe is required in the installed system to make sure the `elemental-system-rancher` is actually capable to reach rancher manager and kick start rancher provisioning.  Then, on rancher side, the proxy needs to be configured so that rancher-system-agent is also capable to see rancher manager. So, up to my understanding the configuration needs to be in three different points:

1. Set `/etc/sysconfig/proxy` in registration yaml:

```yaml
cloud-config:
  write_files:
    - path: /etc/sysconfig/proxy
      owner: root:root
      permissions: '0644'
      content: |
        PROXY_ENABLED="yes"
        HTTP_PROXY="http://192.168.0.1:3128/"
        HTTPS_PROXY="http://192.168.0.1:3128/"
        FTP_PROXY="http://192.168.0.1:3128/"
        NO_PROXY="localhost, 127.0.0.1"
```

2. Set the same exact `/etc/sysconfig/proxy` in seed-image yaml.
3. Set the proxy variables in cluster setup on Rancher side

juadk:
> 1. Set `/etc/sysconfig/proxy` in registration yaml:
> 
> ```yaml
> cloud-config:
>   write_files:
>     - path: /etc/sysconfig/proxy
>       owner: root:root
>       permissions: '0644'
>       content: |
>         PROXY_ENABLED="yes"
>         HTTP_PROXY="http://192.168.0.1:3128/"
>         HTTPS_PROXY="http://192.168.0.1:3128/"
>         FTP_PROXY="http://192.168.0.1:3128/"
>         NO_PROXY="localhost, 127.0.0.1"
> ```
> 
> 2. Set the same exact `/etc/sysconfig/proxy` in seed-image yaml.
> 3. Set the proxy variables in cluster setup on Rancher side

yes, three places to configure the proxy settings but I don't think we can do simpler...

kkaempf (approved):
:+1:

kkaempf:
How do we get that to `rancher-system-agent` ?

juadk:
> How do we get that to `rancher-system-agent` ?

From the UI, with this:
image

PR #797, opened 2023-04-19
Add containerized ISO image

davidcassany:
This is the Dockerfile to build the ISO as a container. It is a natural evolution of the `build-iso` package in OBS and `build-iso` should simply pull this Dockerfile from github. This way the image OS and its ISO are both defined and described within the same place.

Part of to rancher/elemental-operator#396


kkaempf:
There are three occurences of `5.3` in the Dockerfile. Can these be a parameter ?

kkaempf (approved):
:+1:

fgiudici (approved):
lgtm!

davidcassany:
> There are three occurences of `5.3` in the Dockerfile. Can these be a parameter ?

Yes it can, it just goes at the cost of making our already complex `_service` file bigger. Ideally we should be capable to pull it from the prjconf so we could build with same sources for 5.3 and 5.4 in different repositories. Doable, the problem is that this means adding another twist in `_service` files and in prjconf. We can/should consider it though.

Probably we can make it a new issue for building against 5.4 from same sources, to we do not have to maintain to projects, just create the appropriate repos in OBS.

PR #809, opened 2023-04-25
ci: update hardened cluster with k8s 1.25

### LINKED ISSUE #801: ci: bump k8s to 1.25 in hardened cluster tests
Follow up of https://github.com/rancher/elemental/pull/798

It will be not so easier bumping kubernetes there because we have to configure PSA instead of PSP.

juadk:
Fix #801


juadk:
However, I have to update Rancher documentation to include  `cattle-elemental-system`  and `cattle-resources-system` in https://ranchermanager.docs.rancher.com/reference-guides/rancher-security/psa-restricted-exemptions

ldevulder (approved):

PR #816, opened 2023-04-28
ci/ui: increase cluster check status timeout

juadk:
Some tests are still failing when cluster is created, it doesn't show active before reaching the timeout, it look like to be a matter of seconds:

At 9:38, the cluster is not active:
image

20 seconds later, the cluster is active:
image

As mentioned in the comment, it the check fails again, let's try to reload the page and check the status.


fgiudici (approved):
good catch!

PR #822, opened 2023-05-03
Do not compare versions from different repositories on upgrades

### LINKED ISSUE #818: Upgrade should not check version on images from different repositories
What if the upgrade script does not check the version when upgrade is happening to an image of a different repository?

Consider, for instance, upgrading from `rancher/elemental-teal/5.3:1.1.4-6.5` to `rancher/elemental-teal-rt/5.3:1.1.4-2.3`. In that case, images are from different streams, it doesn't make much sense checking for the version as there is no reason to assume there are related.

Same applies for images from stable vs staging or dev, IMHO upgrade should just happen when moving across repositories.

The counter part is that if two different upgrades for different repositories are launched over the same cluster concurrently the result could be inconsistent (some nodes in repo1 and other in repo2)...

davidcassany:
Fixes #818

### DIFF
--- a/framework/files/usr/sbin/suc-upgrade
+++ b/framework/files/usr/sbin/suc-upgrade
@@ -24,13 +24,19 @@ function config()
 function isHigherVersion()
 {
     local img_ver=$( . ${RELEASE_FILE} && printf '%s\n' "${IMAGE_TAG}" )
+    local img_repo=$( . ${RELEASE_FILE} && printf '%s\n' "${IMAGE_REPO}" )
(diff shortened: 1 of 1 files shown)


frelon (approved):
Very nice!

davidcassany:
This turns to be useful to move across different code streams, for instance moving from regular OS to RT versions of it. Or even to move form stable to staging to dev in any direction. It does not have much sense comparing versions for streams. This assumes different repositories in the registry are not necessarily connected in any way, hence versioning could be also radically different and disconnected.

fgiudici (approved):

PR #823, opened 2023-05-03
Bump go-version and deps

frelon:


juadk (approved):

ldevulder:
@frelon why not simply modifying the Go version to bump to 1.20 without adding all the deps? In fact I removed these deps from the E2E tests because it was painful to update this each time we wanted to bump to the latest versions of Ginkgo/Gomega. The idea was to always use the latest versions as well as for the Elemental helpers functions in `ele-testhelpers`.

Issue #634, opened 2023-01-24
Reconsider the use of linuxkit for cloud data providers


davidcassany:
Currently we are using a fork of linuxkit in the shape of a yip plugin. We mostly have fork for two reasons:

* In our fork we include https://github.com/linuxkit/linuxkit/pull/3585
* Our fork has the providers refactored in a package so other tools (aka yip) can use it as a library

The fork was meant to be a temporary fill the gap solution waiting for the azure support to be merged. Our intention was to wait for the azure support upstream merge and then one a PR with our refactor to make it accessible as a library. First step never happened and hence we are still using the fork and the upstream PR of our changes was never done.

This card is about reconsidering linuxkit use and situation. From one side the current fork is under my own user org (davidcassany) and this should not be the case. On an other side the linuxkit project does not seam to be very active and it is unclear if attempting to merge upstream our required changes is a real option. 

* We could simply move our fork to rancher-sandbox and maintain our own
* We could attempt to open a PR with our refactor upstream and if merged move to upstream (no azure support though)
* We could explore other libraries or implementations (afterburn or similar ones)

kkaempf:
While upstream linuxkit appears to be active, they're certainly overwhelmed with issue reports (302 open) and PRs (41 open) with no substantial change in sight.

Otoh, Elemental doesn't currently target Azure so we could live without the Azure provider for the time being (and thus just use upstream linuxkit).

I'd keep the current status quo for now, move from davidcassany/linuxkit to rancher-sandbox/linuxkit, and maintain our own fork.
At the end of the day, it's a `yip` problem and (yip) upstream might find a different solution.

PR #824, opened 2023-05-03
Enable cloud-config from removable devices

davidcassany:
If cloud-init data is provided as the `user-data` file in an eligible device this data is copied to both `/oem/userdata.yaml` and `/oem/userdata`. Since `/oem` is an analyzed path for custom cloud-init files this will be taken into accound on later `elemental run-stage` calls.

### DIFF
--- a/framework/files/system/oem/02_datasources.yaml
+++ b/framework/files/system/oem/02_datasources.yaml
@@ -1,0 +1,6 @@
+name: "Elemental cloud-config from a removable device"
+stages:
+  initramfs:
+  - datasource:
+      providers: ["cdrom"]
+      path: "/oem"


davidcassany:
I did some tests and I could not manage to get a Fat32 formatted USB stick to work... 😞 I seamlessly worked by using an ISO and attaching it as a cdrom to the image on KVM, but it did not make the trick from a USB.

So far I created an iso simply and only incuding a `user-data` file with the following contents:

```yaml
#cloud-config
write_files:
- path: /etc/sysconfig/proxy
  append: true
  content: |
    PROXY_ENABLED="yes"
    HTTP_PROXY=http://some.domain.org:8080
    HTTPS_PROXY=https://some.domain.org:8080
    NO_PROXY="localhost, 127.0.0.1"
```
And it worked fine. The ISO was done with 

```bash
mkisofs -o cidata.iso -V CIDATA -J -r user-data
```
being executed on the folder containing the `user-data` file. Note the `CIDATA` label is mandatory, this is what datasource plugin uses to consider eligible cloud-init devices. This worked fine and it was processed at Live ISO boot and also when booting the installed system. I am exploring why the FAT32 device with `CIDATA` label I created did not work.

davidcassany:
I managed to make use of this feature by fixing the underlying datasource provider in our fork of linuxkit required changes are here https://github.com/davidcassany/linuxkit/pull/7. After merging this yip dependency should be bumped and then bump elemental-cli dependency on yip.

I am starting to believe we should definitely fork the `pkg/metadata` from linuxkit org to `rancher-sandbox` and tie yip there. So we can simply maintain this datasource as a stand alone project. This is related to https://github.com/rancher/elemental/issues/634

davidcassany:
Blocked on https://github.com/davidcassany/linuxkit/pull/7 and eventually on https://github.com/rancher/elemental/issues/634

kkaempf:
we will fork linuxkit to rancher-sandbox

kkaempf:
> we will fork linuxkit to rancher-sandbox

Done. @davidcassany has admin rights 😉

davidcassany:
Default branch of the fork has been renamed to `main` and copied to another branch called `linuxkit_original_master` to keep original branch somewhere. Then a PR against `main` has been opened including our changes and dropping all the code we are not using see https://github.com/rancher-sandbox/linuxkit/pull/1

davidcassany:
This PR is ready to be merged, the changes required to use vFAT filesystems are already merged in rancher-sandbox/linuxkit and a PR to yip to switch to such a project is already opened too https://github.com/mudler/yip/pull/90. The code change here is not dependent on the other PRs, this is already functional for ISOs. As soon as element-cli is updated with new yip it will also support vFAT filesystems.

kkaempf (approved):
🤩

PR #828, opened 2023-05-10
Ensure names are unique for all stages

### LINKED ISSUE #829: Recovery system does not boot
Reboot installed node into recovery partition fails it does not boot, it gets stuck at boot. The `/run/cos/cos-layout.env` file was not created as expected by the cloud-config.

It successfully boots.

IMHO this issue unveils three aspects:

* cloud-config failure/unexpected behavior
* without `/run/cos/cos-layout.env` defaults should have been applied allowing the OS to boot
* What is the expected behavior of the boot assessment when recovery does not boot? Should we retry on active (I doubt it). Simply reboot and cause a reboot loop should not happen, so there isn't a good and simple answer I guess.

**Environment:**

- Elemental release version (use `cat /etc/os-release`): Stable 😞

davidcassany:
We are suffering from a small yip regression. Since v1.0.0 yip uses a DAG (Directed Acyclic Graph) to _sort_ and relate all the required executions, allowing parallel executions (we have not played with it so far). However one of the side effects is that the dag structures items in a map internally using the step name if provided, hence having two steps with the same name causes the step overwrite.

Exactly this is what is happening in the layout setup, we have two steps for that (recovery and non recovery) with the same name, so one of the two gets ignored...

Fixes #829

### DIFF
--- a/framework/files/system/oem/01_elemental-rootfs.yaml
+++ b/framework/files/system/oem/01_elemental-rootfs.yaml
@@ -51,7 +51,7 @@ stages:
         PERSISTENT_STATE_BIND: "true"
     - if: '[ -f "/run/cos/recovery_mode" ]'
       # omit the persistent partition on recovery mode
-      name: "Layout configuration"
+      name: "Layout configuration for recovery"
       environment_file: /run/cos/cos-layout.env
       environment:
         OVERLAY: "tmpfs:25%"


kkaempf (approved):
lgtm

davidcassany:
> Shouldn't this be detected by yip (iow isn't this a yip bug) ?

Yes, this is a yip issue. So this is a simple, sane and fully functional workaround for now.

davidcassany:
Already opened the yip counter part https://github.com/mudler/yip/pull/88

PR #850, opened 2023-05-26
ci: install CRDs chart for operator when needed

ldevulder:
Elemental Operator chart has been split, so newer versions should install all charts.

Verification runs:
- [CLI-K3s-OBS_Dev](https://github.com/rancher/elemental/actions/runs/5092311471)
- [UI-K3s-OBS_Dev](https://github.com/rancher/elemental/actions/runs/5091386989)
- [UI-K3s-OBS_Stable](https://github.com/rancher/elemental/actions/runs/5091802366)

### DIFF
--- a/tests/e2e/upgrade_test.go
+++ b/tests/e2e/upgrade_test.go
@@ -16,6 +16,7 @@ package e2e_test
 
 import (
 	"os"
+	"os/exec"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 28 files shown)


tests/e2e/upgrade_test.go
```diff
@@ -37,17 +37,19 @@ var _ = Describe("E2E - Upgrading Elemental Operator", Label("upgrade-operator")
 	}
 
 	It("Upgrade operator", func() {
-		err := kubectl.RunHelmBinaryWithCustomErr("upgrade", "--install", "elemental-operator",
-			operatorUpgrade,
-			"--namespace", "cattle-elemental-system",
-			"--create-namespace",
-		)
-		Expect(err).To(Not(HaveOccurred()))
+		for _, chart := range []string{"elemental-operator-crds", "elemental-operator"} {
```
davidcassany, line 40: This is the correct approach generally speaking and this is what we should stick with in the future. However this is what is likely to fail when upgrading from current stable to current dev. As the order should be inverted (just for the upgrade from one chart to any other release based on two charts).

Probably here the trick could be similar to the one before, if there is a previous crds chart installation use `[]string{"elemental-operator-crds", "elemental-operator"}` if there is no previous crds chart installation lets use `[]string{"elemental-operator", "elemental-operator-crds"}`.

Anyway I'd leave it for a follow up PR.

davidcassany (approved):
Nice! Looks good to me :+1:

fgiudici (approved):
lgtm

davidcassany (approved):
LGTM

davidcassany:
@ldevulder I saw verification runs are green, with your permission I am merging this PR. Thanks much for the quick solution 🙏🏽

PR #857, opened 2023-06-06
ci/ui: add fully UI configurable workflow

### LINKED ISSUE #834: e2e/ui: Create a new fully customizable manual workflow
We need a workflow like we have for the CLI: https://github.com/rancher/elemental/blob/main/.github/workflows/cli-obs-manual-workflow.yaml

The idea is to be able to give a custom ISO and a custom helmchart for the operator.
However, that's not possible to do it right now because we can only build an ISO from stable in the UI.

juadk:
Fix #834


.github/workflows/ui-obs-manual-workflow.yaml
```diff
@@ -0,0 +1,62 @@
...
+        description: Rancher Manager channel/version to use for installation
+        default: stable/latest
+        type: string
+      runner_template:
+        description: Runner template to use
+        default: elemental-e2e-ci-runner-spot-x86-64-template-n2-standard-16-v4
+        type: string
+
+jobs:
+  cli:
```
ldevulder: `ui` I think instead of `cli` ;-)

ldevulder (changes requested):

ldevulder (approved):

PR #860, opened 2023-06-07
ci/ui: make sure we are on the extension screen

### LINKED ISSUE #859: ci: fix sporadic issue when enabling extension in rancher manager
Test fails when it tries to enable extension support.

image

What happens is we don't give enough time to switch screen or check that we are and the correct screen so sometimes `get button` is applied on the previous screen. The fix will be to check that we are in the correct screen before getting buttons.

In this screenshot, we can see that the `get button` is done on the wrong screen.
Screenshot 2023-06-07 at 09 50 21

juadk:
Fix #859 

## Verification run
image

[UI-K3s-Rancher_Latest](https://github.com/rancher/elemental/actions/runs/5197701991) ✔️ 
[UI-RKE2-Rancher_Latest](https://github.com/rancher/elemental/actions/runs/5197703014) ✔️

### DIFF
--- a/tests/cypress/latest/e2e/unit_tests/elemental_plugin.spec.ts
+++ b/tests/cypress/latest/e2e/unit_tests/elemental_plugin.spec.ts
@@ -41,6 +41,8 @@ filterTests(['main', 'upgrade'], () => {
       topLevelMenu.openIfClosed();
       cy.contains('Extensions')
         .click();
+      // Make sure we are on the Extensions page
+      cy.contains('.message-icon', 'Extension support is not enabled');
       cy.clickButton('Enable');
       cy.contains('Enable Extension Support?')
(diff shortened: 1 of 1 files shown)


tests/cypress/latest/e2e/unit_tests/elemental_plugin.spec.ts
```diff
@@ -41,6 +41,9 @@ filterTests(['main', 'upgrade'], () => {
       topLevelMenu.openIfClosed();
       cy.contains('Extensions')
         .click();
+      // Make sure we are on the Extensions page
+      // https://github.com/rancher/elemental/issues/859
```
ldevulder: This link is not needed in the code, the issue is already linked to this PR.

ldevulder (changes requested):

ldevulder (approved):

PR #874, opened 2023-06-16
Use the new init command

frelon:
In elemental-toolkit v0.11.0 we introduce the init command that initializes configuration for a new derivative.

For elemental-teal we don't need the cloud-config-defaults feature.


davidcassany:
Part of rancher/elemental-toolkit#1792

davidcassany (approved):
LGTM

@frelon after merging verify OBS builds and then pass [CLI-K3s-OBS_Dev](https://github.com/rancher/elemental/actions/workflows/cli-k3s-obs_dev.yaml), this workflow is triggered manually and default values are already good.

PR #890, opened 2023-06-30
Run elemental register at each boot

### LINKED ISSUE #434: Elemental hosts should run elemental-register at each boot
After registration and provisioning, each host should re-register at each reboot.
This would ensure:
1. the host is tracked by a MachineInventory in the Elemental Operator
2. labels and annotations (in particular the IP addr annotation, SMBIOS and HW labels) are updated

Note that point 1. could be useful only if the host has not been made part of a cluster yet, otherwise  the tracked state could be a bit tricky: if the host is already part of a cluster (provisioned with a kubernetes deployment) it cannot be completely re-synced.
This case is *not* part of this card, and should be tracked in a follow-up one.

anmazzotti:
Close rancher/elemental-operator#434


kkaempf:
How does that work for air-gapped systems ?
(Extreme example: a submarine which only comes into port after some months 😉 )

I think we need to differentiate between "first time boot" where a connection to the management cluster should be mandatory and "later boots".

anmazzotti:
> How does that work for air-gapped systems ? (Extreme example: a submarine which only comes into port after some months wink )
> 
> I think we need to differentiate between "first time boot" where a connection to the management cluster should be mandatory and "later boots".

Currently the elemental-register command will endless retry registration. 
With @fgiudici we discussed whether the registration should be retried when not on live system. For example the scenario where there is no registration at all. 

So currently I implemented a way to only retry if we are on a live system. Installed system will try registration only once at each boot, but abort if it fails. 

I am going to open a new issue to discuss how/when we should retry registration and we can try to cover all possible scenarios. 
For example one idea is to define a [StartLimitIntervalSec and StartLimitBurst](https://www.freedesktop.org/software/systemd/man/systemd.unit.html#StartLimitIntervalSec=interval) so that we can keep it consistent and avoid the need to differentiate between first time boot or later boots.

fgiudici (approved):
LGTM!
Anyway, we need https://github.com/rancher/elemental-operator/pull/470 in before merging this: added `blocked` label

PR #891, opened 2023-06-30
Move _constraints into Dockerfile

### LINKED ISSUE #889: _constraints via _service don't work in OBS
[A clear and concise description of what the bug is.]

There are always OBS builds failing with "no space left on device"
We already have a [_constraints file](https://github.com/rancher/elemental/blob/main/.obs/dockerfile/teal-iso/_constraints) asking for 24G of disk. But still :-/

Smooth builds 😆 

[Miscellaneous information that will assist in solving the issue.]

Reaching out to buildservice team, it turns out that _constraints via _service is not supported.

But they merged a fix last week to support
```
#!BuildConstraint: hardware:disk:size unit=G 24
```
within a Dockerfile.

(Not sure if this also work for Dockerfiles downloaded via a _service 🤷🏻 )

kkaempf:
Since OBS doesn't support `_constraints` [via _service](https://github.com/rancher/elemental/issues/889), move them into the Dockerfile.
Constraints in Dockerfile are possible since mid-June 2023.

Fixes #889


fgiudici (approved):
great!

PR #896, opened 2023-07-04
Switch supportlevel to "l3"

### LINKED ISSUE #895: LABEL com.suse.supportlevel="techpreview"
After some searching around confluence I am not sure we should not use this label at all.

This is currently overwriting the support level from SLE Micro and I believe this should match SLE support levels, but as far as I understand we are part of Rancher, hence we should match Rancher support level, don't we? So probably we should should use different label (`com.suse.rancher.supportlevel`?) matching support level scheme of rancher.

I also saw the image `rancher/rancher:v2.7.x` includes:
```
"com.suse.bci.base.supportlevel": "l3",
"com.suse.sle.base.supportlevel": "l3",
"com.suse.supportlevel": "l3"
```

And I believe all of those are inherited from the base BCI image being used. Probably we should also see the support level for the base images specifically? Currently this is not labeled under micro namespace.

-----

After researching in Confluence, this label is in use and should be set to `l3` (== fully supported) for Elemental Teal

Not sure about the `builder-image` though, we might want to declare it as `unsupported` 🤔

kkaempf:
This in accordance to the respective rules documented in Confluence.

At the end of the day, the operating system is BCI based and hence should use the respective supportlevel label values.

Fixes #895


fgiudici (approved):
LGTM

davidcassany (approved):

PR #902, opened 2023-07-06
suc-upgrade: allow OS upgrades on hosts with failed systemd units

### LINKED ISSUE #901: OS upgrades not allowed when `systemctl is-system-running` reports a `degraded` state
On provisioned hosts, if there a systemd service fails causing `systemctl is-system-running` to report a `degraded` state OS updates are skipped.

OS upgrades should be allowed also when systemd reports a degraded state: they could even fix the issue.

Issue is in the suc-upgrade script here:
https://github.com/rancher/elemental/blob/b73cb4fb082aea15b5d056acea9a54dc5b59e548/framework/files/usr/sbin/suc-upgrade#L62

fgiudici:
Before update we check the system status as reported by systemd via the `systemctl is-system-running` cmd, as we want to skip in case the system is shutting down. This would prevent overlapping OS updates.
We rely on the command exit code only, which [returns error (EXIT_FAILURE) when the state is anything but "running"](https://github.com/systemd/systemd/blob/0a3d108f46c39ada3e4db954f65d8e2fdab4f6b5/src/systemctl/systemctl-is-system-running.c#L24).

The [available states](https://github.com/systemd/systemd/blob/5bc9ea070f7734a9b86c6f7f8e2ed4365229a6c1/src/core/manager.c#L4857) are:
- "initializing"
- "starting"
- "running"
- "degraded"
- "maintenance"
- "stopping"

This patch enables OS upgrades also when the system is in the "degraded" state.
We have to prevent OS upgrades for the "stopping" state.
For the other states it could be arguable whether it makes sense to hold the OS upgrade or not: this patch maintains the previous behavior.

NOTE: fixed also the tab spacing while there. Anyway, kept in a separate commit, actual patch is pretty small.

Fixes #901

### DIFF
--- a/framework/files/usr/sbin/suc-upgrade
+++ b/framework/files/usr/sbin/suc-upgrade
@@ -46,63 +46,63 @@ function isHigherVersion()
 }
 
 function isEqualVersion() {
- if diff $RELEASE_FILE ${HOST_DIR}${RELEASE_FILE} >/dev/null; then
-     return 0
- fi
- return 1
+    if diff $RELEASE_FILE ${HOST_DIR}${RELEASE_FILE} >/dev/null; then
+        return 0
+    fi
+    return 1
 }
 
 function isActiveSystem() {
-  [ ! -f "${ACTIVE_FILE}" ] && return 1
-  return 0
+    [ ! -f "${ACTIVE_FILE}" ] && return 1
+    return 0
 }
 
 (
-  flock -w $LOCK_TIMEOUT 200 || exit 1
-  if ! nsenter -i -m -t 1 -- systemctl is-system-running; then
-      # Exit if there is a shutdown process already going on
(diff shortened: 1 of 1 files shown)


kkaempf (approved):
🤞🏻

kkaempf:
How can we improve testing here ?

/cc @juadk @ldevulder

davidcassany (approved):
LGTM

davidcassany:
Just for further background, these sort of checks within script they are all there to prevent concurrent upgrades. IMHO this should be handled at System Upgrade Controller level, there is a stalled issue about it here https://github.com/rancher/system-upgrade-controller/issues/225 . If no concurrent upgrades are prevented at a higher level I'd say we can get rid of some the checks within the upgrade script, like the one causing the issue here.

PR #905, opened 2023-07-07
FORCE flag skips all sanity checks on upgrades

davidcassany:
This commit ensure FORCE flag on upgrades skips all sanity checks. In addition, this commit also relaxes the systemd state check to only abort upgrades on stopping state.

Related to #901

### DIFF
--- a/framework/files/usr/sbin/suc-upgrade
+++ b/framework/files/usr/sbin/suc-upgrade
@@ -59,27 +59,28 @@ function isActiveSystem() {
 
 (
     flock -w $LOCK_TIMEOUT 200 || exit 1
-    if ! SYSSTATUS=`nsenter -i -m -t 1 -- systemctl is-system-running`; then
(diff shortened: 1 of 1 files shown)


framework/files/usr/sbin/suc-upgrade
```diff
@@ -59,27 +59,28 @@ function isActiveSystem() {
...
+        *)
+            # other states (degraded, maintenance...) should not stop OS upgrades, see https://github.com/rancher/elemental/issues/901
+            ;;
         esac
     fi
 
-    if isEqualVersion; then
-        echo "Upgrade already done with release:"
-        cat ${HOST_DIR}${RELEASE_FILE}
+    if [ "$FORCE" != "true" ]; then
```
fgiudici, line 75: good catch 👍🏼

fgiudici (approved):
LGTM

PR #906, opened 2023-07-10
Improve registration on already installed systems

### LINKED ISSUE #471: Improve re-registration behavior on already installed systems
This is a follow up of #434 

In that issue we enabled re-registration at each boot, however:
- We only try once and in case of failure we abort
- We do not handle the "multiple machines re-booting all at the same time" scenarios, so in some context with hundreds of devices it won't scale well
- We are making a distinction between "live system" and "installed system", which adds complexity. Ideally we should retry consistently without having to think about the installation state

One possible solution is to actually remove logic from the `elemental-register` code and rely on systemd instead, for example by defining [StartLimitIntervalSec and StartLimitBurst](https://www.freedesktop.org/software/systemd/man/systemd.unit.html#StartLimitIntervalSec=interval).  

This however won't really cover the "submarine coming to surface after months" scenario.  
For that we may need to actually implement an exponential backoff so that we can keep trying (at max once every 30 min?) indefinitely.

anmazzotti:
Resolve rancher/elemental-operator#471

### DIFF
--- a/.obs/specfile/elemental.spec
+++ b/.obs/specfile/elemental.spec
@@ -84,19 +84,19 @@ rm -rf %{buildroot}/usr/libexec/.placeholder
 %service_add_post elemental-populate-node-labels.service
 %service_add_post shutdown-containerd.service
 %service_add_post elemental-register.service
-%service_add_pre elemental-register.timer
+%service_add_post elemental-register.timer
 
 %preun
 %service_del_preun elemental-populate-node-labels.service
 %service_del_preun shutdown-containerd.service
 %service_del_preun elemental-register.service
-%service_add_pre elemental-register.timer
+%service_del_preun elemental-register.timer
 
 %postun
 %service_del_postun elemental-populate-node-labels.service
 %service_del_postun shutdown-containerd.service
 %service_del_postun elemental-register.service
-%service_add_pre elemental-register.timer
+%service_del_postun elemental-register.timer
 
 %files
 %defattr(-,root,root,-)
--- a/.obs/specfile/elemental.spec
+++ b/.obs/specfile/elemental.spec
@@ -78,21 +78,25 @@ rm -rf %{buildroot}/usr/libexec/.placeholder
 %service_add_pre elemental-populate-node-labels.service
 %service_add_pre shutdown-containerd.service
 %service_add_pre elemental-register.service
+%service_add_pre elemental-register.timer
 
 %post
 %service_add_post elemental-populate-node-labels.service
 %service_add_post shutdown-containerd.service
 %service_add_post elemental-register.service
+%service_add_pre elemental-register.timer
 
 %preun
 %service_del_preun elemental-populate-node-labels.service
 %service_del_preun shutdown-containerd.service
 %service_del_preun elemental-register.service
+%service_add_pre elemental-register.timer
 
 %postun
 %service_del_postun elemental-populate-node-labels.service
 %service_del_postun shutdown-containerd.service
 %service_del_postun elemental-register.service
+%service_add_pre elemental-register.timer
 
 %files
 %defattr(-,root,root,-)
@@ -109,6 +113,7 @@ rm -rf %{buildroot}/usr/libexec/.placeholder
 %dir %{_unitdir}
 %{_unitdir}/shutdown-containerd.service
 %{_unitdir}/elemental-register.service
+%{_unitdir}/elemental-register.timer
 %{_unitdir}/elemental-populate-node-labels.service
 %{_sbindir}/elemental-populate-node-labels
 %dir /usr/libexec
--- a/framework/files/system/oem/99_elemental-register.yaml
+++ b/framework/files/system/oem/99_elemental-register.yaml
@@ -2,4 +2,4 @@ name: "Elemental operator bootstrap"
 stages:
   network.after:
     - commands:
-        - systemctl start elemental-register.service
+        - systemctl start elemental-register.timer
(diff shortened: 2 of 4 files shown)


framework/files/system/oem/99_elemental-register.yaml
```diff
@@ -2,4 +2,4 @@ name: "Elemental operator bootstrap"
 stages:
   network.after:
     - commands:
-        - systemctl start elemental-register.service
+        - systemctl start elemental-register.timer
```
davidcassany, line 5: I think we can consider entirely removing this file and start the registration timer by default. The default enabled services are listed in a systemd Elemental branding package in [OBS](https://build.opensuse.org/package/show/isv:Rancher:Elemental:Dev/systemd-presets-branding-Elemental).

.obs/specfile/elemental.spec
```diff
@@ -78,21 +78,25 @@ rm -rf %{buildroot}/usr/libexec/.placeholder
 %service_add_pre elemental-populate-node-labels.service
 %service_add_pre shutdown-containerd.service
 %service_add_pre elemental-register.service
+%service_add_pre elemental-register.timer
 
 %post
 %service_add_post elemental-populate-node-labels.service
 %service_add_post shutdown-containerd.service
 %service_add_post elemental-register.service
+%service_add_pre elemental-register.timer
```
davidcassany: should be `%service_add_post`. Copy & paste error I guess.

.obs/specfile/elemental.spec
```diff
@@ -78,21 +78,25 @@ rm -rf %{buildroot}/usr/libexec/.placeholder
...
 
 %preun
 %service_del_preun elemental-populate-node-labels.service
 %service_del_preun shutdown-containerd.service
 %service_del_preun elemental-register.service
+%service_add_pre elemental-register.timer
```
davidcassany: Same sort error here as above, the macro is not correct

.obs/specfile/elemental.spec
```diff
@@ -78,21 +78,25 @@ rm -rf %{buildroot}/usr/libexec/.placeholder
...
 
 %postun
 %service_del_postun elemental-populate-node-labels.service
 %service_del_postun shutdown-containerd.service
 %service_del_postun elemental-register.service
+%service_add_pre elemental-register.timer
```
davidcassany: Same sort error here as above, the macro is not correct

framework/files/usr/lib/systemd/system/elemental-register.service
```diff
@@ -3,10 +3,19 @@ Description=Elemental Register
 Documentation=https://elemental.docs.rancher.com
 Wants=network-online.target
 After=network-online.target
+# Backoff after 5 attempts in 5 min 
+StartLimitIntervalSec=5min
```
davidcassany, line 7: Out of curiosity. Do we know if once the limit is achieved the timer will restart the counter?

davidcassany (changes requested):
Looks good just some copy & paste errors and I wonder if we shouldn't set the timer enabled by default and not include any cloud-config for it at all.

framework/files/usr/lib/systemd/system/elemental-register.service
anmazzotti: > Out of curiosity. Do we know if once the limit is achieved the timer will restart the counter?

Yes, this is actually the reason we use `CollectMode=inactive-or-failed`, so that it will not enter a `failed` state and the Timer will just restart it.

.obs/specfile/elemental.spec
anmazzotti: Definitely too much copy/paste. Thank you.

framework/files/usr/lib/systemd/system/elemental-register.timer
```diff
@@ -0,0 +1,15 @@
+[Unit]
+Description=Run elemental-register every 30 minutes
+Documentation=https://elemental.docs.rancher.com
+
+[Timer]
+OnStartupSec=5
```
davidcassany, line 6: I am wondering if this shouldn't we use the `OnBootSec` here, but not so sure to be honest. It is not clear to me to what point this delay is applied to. Is it when the timer is started? So it starts the registration after 5seconds the timer starts? If that's the case then this is fine I'd say.

framework/files/system/oem/99_elemental-register.yaml
anmazzotti: I would leave it here right now for simplicity, this issue already has three PRs as dependencies. 
I can open a follow up issue to discuss this if you think it's the case.

davidcassany: Yes, in fact. Also thinking it further probably we can keep it like this as running the service from the cloud-config allows us to set a different start or setup for different scenarios (live media, active, fallback or recovery systems).

We could omit it in live media and in that case run something more aggressive like `elemental-register --install` (assuming we have a flag to trigger installation or not)

davidcassany (approved):
Looks good to me

framework/files/usr/lib/systemd/system/elemental-register.timer
anmazzotti: It should be equivalent for Timers as far I can understand the [doc](https://www.freedesktop.org/software/systemd/man/systemd.timer.html#Options).

```text
Defines a timer relative to when the service manager was first started.  
For system timer units this is very similar to OnBootSec= as the system service manager is generally started very early at boot.  
It's primarily useful when configured in units running in the per-user service manager, as the user service manager is generally started on first login only, not already during boot.
```

I thought OnStartup would be safer as we have dependencies on network and so on.

Issue #815, opened 2023-04-26
Update to SLE Micro for Rancher 5.4


kkaempf:
**Describe the solution you'd like:**

Update from Elemental Teal 5.3 to Elemental Teal 5.4

Make Elemental Teal 5.4 the default.

**Anything else you would like to add:**

`suse/sle-micro-rancher/5.4` is now available in OBS.
We should update soon.

davidcassany:
I am already consistently building for 5.3 and 5.4 (next repositories) within the same project [here](https://build.opensuse.org/project/show/home:dcassany:ElementalNewDev). 

~I managed to simplify the image `Dockerfile` and its `_service` files, thanks an undocumented feature about passing `ARG`s to `Dockerfile` by using the OBS config like  `BuildFlags: dockerarg:SLEMICRO_VERSION=5.4`. I could not manage to do the same for charts, but given we might consider moving them form OBS to github pages as part of rancher/elemental-operator#460 I won't spend time on that. Gonna try to learn how to release charts the way we need via github pages.~

~Both issues (current one and rancher/elemental-operator#460), even being technically unrelated, should be considered/merged together as both have implications on the CI, so I'd rather try to collect everything together and apply the changes once. Trying to avoid a back and forth changes hitting the CI.~

Edit: At the end I kept the SLE Micro version reference within image reference as we need the `latest` tag for testing and, as @kkaempf pointed, using `latest` tag without distinguishing the underlying OS stream will be problematic for releasing OS rebuilds if needed.

davidcassany:
I managed to pass the CI against my repository using the builds based on 5.4 [here](https://github.com/rancher/elemental/actions/runs/5716448119) I am seeing the PRs ready to be reviewed. On merge they will cause breaks in Dev as a new OBS setup will be required.

kkaempf:
>  On merge they will cause breaks in Dev as a new OBS setup will be required.

To be expected. Go for it 😉

davidcassany:
I considered it done, Dev and Staging are already in SLE Micro 5.4 and smoke tests are passing. There are some glitches with some UI tests and upgrade tests are to be validated within the following CI schedule, I could not trigger them manually.

PR #914, opened 2023-07-14
Drop SLE Micro version references from code

davidcassany:
These are the changes required in this repository for the builds I prepared in [here](https://build.opensuse.org/project/show/home:dcassany:ElementalNewDev). 

Note this changes image and charts URLs, so this is likely to be a breaking change from CI perspective.

Part of #815

### DIFF
--- a/.obs/dockerfile/teal-iso/Dockerfile
+++ b/.obs/dockerfile/teal-iso/Dockerfile
@@ -1,5 +1,4 @@
 # SPDX-License-Identifier: Apache-2.0
-#!BuildTag: rancher/elemental-teal-iso:latest
 #!BuildTag: rancher/elemental-teal-iso:%VERSION%
 #!BuildTag: rancher/elemental-teal-iso:%VERSION%-%RELEASE%
 #!BuildConstraint: hardware:disk:size unit=G 10
--- a/.obs/dockerfile/teal-iso/Dockerfile
+++ b/.obs/dockerfile/teal-iso/Dockerfile
@@ -1,12 +1,14 @@
 # SPDX-License-Identifier: Apache-2.0
-#!BuildTag: rancher/elemental-teal-iso/5.3:latest
-#!BuildTag: rancher/elemental-teal-iso/5.3:%VERSION%
-#!BuildTag: rancher/elemental-teal-iso/5.3:%VERSION%-%RELEASE%
+#!BuildTag: rancher/elemental-teal-iso:latest
+#!BuildTag: rancher/elemental-teal-iso:%VERSION%
+#!BuildTag: rancher/elemental-teal-iso:%VERSION%-%RELEASE%
 #!BuildConstraint: hardware:disk:size unit=G 10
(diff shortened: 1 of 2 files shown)


.obs/dockerfile/teal-iso/Dockerfile
```diff
@@ -1,12 +1,14 @@
 # SPDX-License-Identifier: Apache-2.0
-#!BuildTag: rancher/elemental-teal-iso/5.3:latest
-#!BuildTag: rancher/elemental-teal-iso/5.3:%VERSION%
-#!BuildTag: rancher/elemental-teal-iso/5.3:%VERSION%-%RELEASE%
+#!BuildTag: rancher/elemental-teal-iso:latest
```
kkaempf: Hmm, this might fail while we build two versions (like `5.3` and `5.4`) in parallel.

We should probably just drop the `:latest` tag ?!

.obs/dockerfile/teal-iso/Dockerfile
```diff
@@ -1,12 +1,14 @@
...
+#!BuildTag: rancher/elemental-teal-iso:%VERSION%
+#!BuildTag: rancher/elemental-teal-iso:%VERSION%-%RELEASE%
 #!BuildConstraint: hardware:disk:size unit=G 10
 #!BuildConstraint: hardware:memory:size unit=G 16
 
-FROM rancher/elemental-teal/5.3:latest AS os
-FROM rancher/elemental-builder-image/5.3:latest AS builder
+ARG SLE_VERSION
+
+FROM rancher/elemental-teal:latest AS os
```
kkaempf: Same here ...

kkaempf (changes requested):

.obs/dockerfile/teal-iso/Dockerfile
davidcassany: In OBS builds this will end up being something like:
```
registry.org/home/cassany/elementalnewdev/containers/rancher/elemental-teal-iso:latest
registry.org/home/cassany/elementalnewdev/containers_next/rancher/elemental-teal-iso:latest
```

So there will not be an overlap as long as they build in different repositories. For IBS builds under registry.suse.com I see your point. If we follow other SUSE images approach we could keep it our latest code base stream build and remove it on old ones. Using `suse/sle15:latest` image as an example this shifted from SP4 to SP5 as soon as the new one was released. I agree this is quite annoying from maintenance point of view as this forces us to update the old build in any new release 🤔 also feels like this can be error prone if done manually.

kkaempf (approved):
Looks much better now 😆

davidcassany:
> Looks much better now laughing

I ended up keeping the `latest` tag and the sle micro version as part of the image repository in registry. We need the latest tag for testing, it will be painful without it, it is just so handy for testing dev. As a counter part I made the sle micro version string dynamic with further `_service` tricks

fgiudici (approved):

PR #916, opened 2023-07-17
Drop /etc/cos/bootargs.cfg from elemental.spec

kkaempf:
It's also packaged in elemental-grub-bootargs (coming from rancher/elemental-toolkit)

### DIFF
--- a/.obs/specfile/elemental.spec
+++ b/.obs/specfile/elemental.spec
@@ -98,8 +98,6 @@ rm -rf %{buildroot}/usr/libexec/.placeholder
 %defattr(-,root,root,-)
 %doc README.md
 %license LICENSE
-%dir %{_sysconfdir}/cos
-%config %{_sysconfdir}/cos/bootargs.cfg
 %dir %{_sysconfdir}/dracut.conf.d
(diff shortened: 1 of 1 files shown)


davidcassany (changes requested):
I don't think this is fully correct, we need this file. The _equivalent_ file from `elemental-grub-bootargs` is not the same and it might not work for k8s provisioning. Note `elemental-grub-bootargs` is not installed in Elemental Teal the `elemental-toolkit` does [not require it](https://github.com/rancher/elemental-toolkit/blob/d1ae3f9a425de2618f9058f3b37583ef3ce52c7d/.obs/specfile/elemental-toolkit.spec#L31-L34).

@kkaempf I am wondering in which context you saw the conflict. Also important to note that all elemental-toolkit subpackages are about to disappear as soon as we align with elemental-toolkit upstream `init` command.

kkaempf:
I saw it when trying to build a [CentOS 8 image](https://build.opensuse.org/package/show/home:kwk:osimage/centos4r) and struggling with `dnf` to add a repo. So instead I added all built rpms and the conflict emerged.

Closing as we're about to drop subpackages.

PR #926, opened 2023-07-19
Drop systemd_pre

kkaempf:
It does not exist :facepalm:

### DIFF
--- a/.obs/specfile/elemental.spec
+++ b/.obs/specfile/elemental.spec
@@ -80,11 +80,6 @@ rm -rf %{buildroot}/usr/libexec/.placeholder
 %service_add_pre shutdown-containerd.service
 %service_add_pre elemental-register.service
 %service_add_pre elemental-register.timer
-%else
-%systemd_pre elemental-populate-node-labels.service
-%systemd_pre shutdown-containerd.service
-%systemd_pre elemental-register.service
-%systemd_pre elemental-register.timer
 %endif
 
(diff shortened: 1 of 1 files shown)


anmazzotti (approved):
This looks good. However I do wonder why we use pre **and** post. Shouldn't `service_add_post` and `service_del_preun` be enough?

Also regarding the `systemd_preun`, I think it actually prevents the restart if I read [this](https://docs.fedoraproject.org/en-US/packaging-guidelines/Scriptlets/#_systemd) doc. I think `systemd_postun_with_restart` should be used instead.

Anyway not related to this PR.

kkaempf:
Well, we just follow the [openSUSE systemd packaging guidelines](https://en.opensuse.org/openSUSE:Systemd_packaging_guidelines) (resp. the [Fedora one's](https://docs.fedoraproject.org/en-US/packaging-guidelines/Scriptlets/#Systemd)) here 🤷🏻

kkaempf:
> Also regarding the `systemd_preun`, I think it actually prevents the restart if I read [this](https://docs.fedoraproject.org/en-US/packaging-guidelines/Scriptlets/#_systemd) doc. I think `systemd_postun_with_restart` should be used instead.

Yes, I thought about this. But since we're not doing package-based updates, this is not relevant and I decided to use `systemd_preun` (which, strictly speaking, wouldn't be needed either).

PR #934, opened 2023-07-27
Separate elemental-register --install

### LINKED ISSUE #526: Remote Reset of Machine when Cluster is deleted or Machine is removed from Pool
When a Cluster is deleted, the MachineInventory gets deleted as well but the node itself doesn't do any clean up and is left in a state where it needs to be re-provisioned by someone with physical access. 

Instead, what I would like to see is that when a cluster gets deleted (or a node is removed from the cluster):

- The MachineInventories don't get deleted 
  - The ownership link would be cleaned up
  - The status for applied plans gets reset and the "default plan" gets queued up again 
- The node is cleaned up using something like `elemental reset --reset-persistent --reset-oem --reboot`
- When the node comes back up, it looks like it was a clean rebuild

Open questions:
- What to do about plans that were included on first boot using the cloud-config? Can we replay those again?
- Should we provide a way to manually reset at any time?

anmazzotti:
Resolves rancher/elemental#526

Create a separate cloud-init to run only from live systems, to exploit the `elemental-register` `--install` argument.

### DIFF
--- a/framework/files/system/oem/05_motd_and_autologin.yaml
+++ b/framework/files/system/oem/05_motd_and_autologin.yaml
@@ -7,7 +7,7 @@ stages:
       content: |
 
         Installation is in progress.
-        You can "journalctl -f -u elemental-register" to view progress
+        You can "journalctl -f -t elemental-register" to view progress
          
       permissions: 0644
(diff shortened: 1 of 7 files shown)


davidcassany (approved):
LGTM, nice :+1:

PR #939, opened 2023-08-01
Update container URLs in CI 

davidcassany:
Part of #815


davidcassany:
We should have way to parametrize the SLE Micro version so we do not need to do all this changes every time we bump the underlaying OS, however I leave it for a separate issue (#940).

fgiudici (approved):
teal53/15.4 --> containers ❤️

Issue #940, opened 2023-08-01
Parametrize Micro version in CI


davidcassany:
**Describe the solution you'd like:**
[A clear and concise description of what you want to happen.]
I'd like to be capable to update the SLE Micro version without having to hunt URL references in the CI (see #939). Ideally we should only change the SLE Micro version value in a single place.

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]
There is a new OS version twice a year, so this is likely to happen relatively often.

**Environment:**

- Elemental release version (use `cat /etc/os-release`):
- Rancher version:
- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:

kkaempf:
no longer relevant

PR #941, opened 2023-08-02
Dynamically set the URL template based on the build env

davidcassany:
A follow up of #939

### DIFF
--- a/tests/cypress/latest/support/commands.ts
+++ b/tests/cypress/latest/support/commands.ts
@@ -388,7 +388,10 @@ Cypress.Commands.add('checkLabelSize', (sizeToCheck) => {
 
 // Add an OS version channel
 Cypress.Commands.add('addOsVersionChannel', (channelVersion) => {
-    const channelRepo = `registry.opensuse.org/isv/rancher/elemental/${channelVersion}/containers/rancher/elemental-teal-channel/5.4:latest`;
+    let channelRepo = `registry.opensuse.org/isv/rancher/elemental/${channelVersion}/containers/rancher/elemental-teal-channel/5.4:latest`;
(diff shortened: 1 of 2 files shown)


tests/cypress/latest/support/commands.ts
```diff
@@ -388,7 +388,10 @@ Cypress.Commands.add('checkLabelSize', (sizeToCheck) => {
 
 // Add an OS version channel
 Cypress.Commands.add('addOsVersionChannel', (channelVersion) => {
-    const channelRepo = `registry.opensuse.org/isv/rancher/elemental/${channelVersion}/containers/rancher/elemental-teal-channel/5.4:latest`;
+    let channelRepo = `registry.opensuse.org/isv/rancher/elemental/${channelVersion}/containers/rancher/elemental-teal-channel/5.4:latest`;
```
kkaempf, line 391: Hmm, this still has an embedded version number (5.4) 🤔

kkaempf (approved):
worth a try 😆

tests/cypress/latest/support/commands.ts
davidcassany: Yes, I already created a CI card to address that here #940 this is a more generic issue within the CI, there are many references out there to the SLE Micro version. They should be limited and parametrized, I'd prefer to have Loic and Juilen around to address this, they might have opinions and good ideas about how to handle this.

davidcassany: Also in relation to that there is also https://github.com/rancher/elemental-operator/pull/491 which is about removing SLE Micro version references from images that are not explicitly based on SLE Micro for rancher (all but the OS images)

PR #942, opened 2023-08-02
Drop the need of the builder image, use the OS itself for that

davidcassany:


kkaempf:
😮

kkaempf (approved):
This surprises me - a lot 😆

davidcassany:
> This surprises me - a lot laughing

Just an idea I had to reduce the images we actually need. I thought that if we include all the packages that are required for building and ISO as part of the SLE Micro for Rancher we can actually use the same image for building. This is a similar trick to what we already do in elemental-toolkit tests. The example OS is already being used to build the ISO of itself. If we do it this way we could drop the builder image, which is a tricky one as it is not based on SLE Micro for Rancher currently.

.obs/dockerfile/teal-os/Dockerfile
```diff
@@ -17,6 +17,7 @@ RUN zypper in -y -- \
   -systemd-presets-branding-SLE-Micro-for-Rancher \
   cryptsetup \
   elemental \
+  mtools \
```
davidcassany: If the test succeeds I'd suggest add it to a requirement to elemental-cli package together with xorriso, this way we ensure any image containing elemental-cli is capable to build ISOs.

davidcassany:
Goes together with https://github.com/rancher/elemental-toolkit/pull/1812

PR #949, opened 2023-08-10
Package elemental-register-install.service

kkaempf:


anmazzotti (approved):
woops, I indeed missed this one. 
great catch, thank you!

PR #951, opened 2023-08-17
Fix elemental-register systemd service

ldevulder:
RemainAfterExit should be defined in Unit section.

This PR should fix this error I saw on node:
Screenshot from 2023-08-17 14-53-23


kkaempf (approved):
😮 
Why didn't we see this before ?!

ldevulder:
> Why didn't we see this before ?!

Because QA was in vacation! :D But in fact it has been added recently.

anmazzotti:
> 😮 Why didn't we see this before ?!

I tested this locally on my machine and I convinced the reviewer it would work. :(
I think I made a copy/paste error though.

I now learned how to build a local iso, so I can test the real thing next time.

kkaempf:
As long as `Staging` isn't affected, I'm relieved 😆

PR #953, opened 2023-08-17
Add SLE-Micro-Rancher Dockerfile for buildservice automation

kkaempf:
We don't have [node-image](https://build.opensuse.org/package/show/isv:Rancher:Elemental:Dev/node-image) when we switch to SLE Micro 5.5 based builds.
It will be replaced by this SLE Micro for Rancher image.

### DIFF
--- a/.obs/dockerfile/slem4r-os/Dockerfile
+++ b/.obs/dockerfile/slem4r-os/Dockerfile
@@ -1,0 +1,97 @@
+# SPDX-License-Identifier: Apache-2.0
+# Define the names/tags of the container
+#!BuildTag: suse/sle-micro-rancher/%%SLEMICRO_VERSION%%:latest
+#!BuildTag: suse/sle-micro-rancher/%%SLEMICRO_VERSION%%:%RELEASE%
+#!BuildConstraint: hardware:disk:size unit=G 8
+#
+
+FROM suse/sle15:15.5 as host
+
+MAINTAINER SUSE LLC (https://www.suse.com/)
+
+# Define labels according to https://en.opensuse.org/Building_derived_containers
+# labelprefix=com.suse.sle.micro.rancher
+LABEL org.opencontainers.image.title="SLE Micro for Rancher"
+LABEL org.opencontainers.image.description="Image containing SLE Micro for Rancher - a containerized OS layer for Kubernetes."
+LABEL org.opencontainers.image.version="%%SLEMICRO_VERSION%%.%RELEASE%"
+LABEL org.opencontainers.image.url="https://www.suse.com/products/micro/"
+LABEL org.opencontainers.image.created="%BUILDTIME%"
+LABEL org.opencontainers.image.vendor="SUSE LLC"
+LABEL org.opencontainers.image.source="%SOURCEURL%"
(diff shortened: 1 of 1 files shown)


.obs/dockerfile/slem4r-os/Dockerfile
```diff
@@ -0,0 +1,97 @@
...
+
+# make ARM happy
+#!ArchExclusiveLine: aarch64
+RUN if [ `uname -m` = "aarch64" ]; then zypper --installroot /osimage in -y raspberrypi-firmware raspberrypi-firmware-config raspberrypi-firmware-dt u-boot-rpiarm64 grub2-arm64-efi; fi
+                
+# make SUSE happy
+RUN zypper --installroot /osimage in --no-recommends -y SLE-Micro-Rancher-release systemd-presets-branding-SLE-Micro-for-Rancher
+
+# make elemental-register happy
+RUN zypper --installroot /osimage in --no-recommends -y dmidecode lvm2
```
anmazzotti, line 53: shouldn't this be part of the elemental-register spec file? I tried to find such file but it does not exist, but still I would expect these to be required dependencies there and to be installed on cascade by installing `elemental`

.obs/dockerfile/slem4r-os/Dockerfile
```diff
@@ -0,0 +1,97 @@
...
+
+# make Rancher (containerd) happy
+RUN zypper --installroot /osimage in --no-recommends -y apparmor-parser
+
+# add elemental
+RUN zypper --installroot /osimage in --no-recommends -y elemental
+
+# make users happy
+RUN zypper --installroot /osimage in --no-recommends -y procps openssl openssh vim-small less iputils kernel-firmware-all NetworkManager-wwan cryptsetup podman zypper
```
anmazzotti, line 62: Is zypper going to work anyway given the read-only filesystem?

anmazzotti (changes requested):

.obs/dockerfile/slem4r-os/Dockerfile
kkaempf: You're probably right. But that might need a longer discussion (with David 😆 )

.obs/dockerfile/slem4r-os/Dockerfile
kkaempf: There's no read-only filesystm ad this point. It's a sub-directory (`/osimage`) in a Docker build 😉

anmazzotti: No I mean why we do install zypper as dependency.
And shouldn't zypper already be included anyway?

anmazzotti: Oh I see it now, so we don't copy the entire `host` container, but you do a selection, this explains why we install the kernel and some other dependencies. So indeed zypper is not going to be included, and you need to run the clean later. 
Ok.

.obs/dockerfile/slem4r-os/Dockerfile
anmazzotti: I'm convinced enough. :D

anmazzotti (approved):

PR #971, opened 2023-08-23
Disable service triggers on staging

davidcassany:
Since all services are set to `manual` or `buildtime` in staging, triggering services has no effect. Staging has to be updated manually.

Related to rancher/elemental-operator#498 and rancher/elemental-toolkit#1817


fgiudici (approved):
👍🏼

PR #978, opened 2023-08-29
Add elemental-register-reset.service

### LINKED ISSUE #977: Elemental reset flow does not retry in case of failures
The last part of the elemental reset flow is prone to failures.
The problem occur when rebooting in recovery mode and executing the `elemental-register --reset` command.

If the command fails (due to a temporary network issue for example), the reset is not reattempted. 

There are conditions where re-attempting is not going to help, for example in case of disk or other hardware failures, but in general it would be more robust to re-attempt indefinitely on the same line of  the `elemental-register-install.service`

Simply creating an equivalent `elemental-register-reset.service` would make the last part of this flow a bit more robust.

anmazzotti:
Resolves #977

### DIFF
--- a/.obs/specfile/elemental.spec
+++ b/.obs/specfile/elemental.spec
@@ -80,6 +80,7 @@ rm -rf %{buildroot}/usr/libexec/.placeholder
 %service_add_pre shutdown-containerd.service
 %service_add_pre elemental-register.service
 %service_add_pre elemental-register-install.service
+%service_add_pre elemental-register-reset.service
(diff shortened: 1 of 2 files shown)


davidcassany (approved):
LGTM

juadk:
Tested successfully with the latest dev operator:
I tested it by unplugging the network cable when the machine rebooted on the recovery partition, I saw the `elemental-register --reset` command running every 5 seconds.
Then, I plugged my network cable back and the magic happened! All good! Thanks

PR #983, opened 2023-08-30
Fix image/iso build

ldevulder:
Image/iso built on CI should be based on the same version as the one built on OBS.

Not perfect but these changes should make the GH artifacts close to the OBS one. This is more a workaround until some changes currently tested by @davidcassany should help us to synchronise Dev build with CI.


davidcassany (approved):
Looks good to me  :+1:

Thanks  for looking at it

PR #1006, opened 2023-09-19
Include system agent unit file in specfile

davidcassany:


davidcassany:
Goes together with https://build.opensuse.org/request/show/1112305

anmazzotti (approved):
Looks good!

PR #1025, opened 2023-09-26
Remove superfluous copy

frelon:
Not needed anymore since https://github.com/rancher/elemental-toolkit/pull/1832


davidcassany (approved):
good catch, even we probably can drop this dockerfile in the near future

frelon:
Hmm, looking at the tests we will have to do a new release of the toolkit. I can revisit this when we get there :+1:

PR #1036, opened 2023-10-05
Remove build of artifacts

ldevulder:
We will no longer build container image and ISO in the GitHub repository. All is now done in OBS/IBS.

Elemental-toolkit repository contains example to build OS image from multiple distributions.

Verification runs:
- [CLI-K3s-OBS_Dev](https://github.com/rancher/elemental/actions/runs/6420726470)
- [CLI-RKE2-OBS_Dev](https://github.com/rancher/elemental/actions/runs/6421076002)
- [UI-K3s-Rancher_Stable](https://github.com/rancher/elemental/actions/runs/6429082084)
- [UI-RKE2-Rancher_Stable](https://github.com/rancher/elemental/actions/runs/6429083896)

NOTE: the issues are not related to this PR and will be investigated/fixed on another issue/PR.


.github/elemental-iso-add-registration
davidcassany, line 1: I would keep this file around. I would like to get rid of this too, but this is being used in several docs articles, hence I'd vote to keep it for the time being until we either improve our docs or we either provide a more refined approach to customize ISOs and/or build custom ISOs.

davidcassany (commented):
@ldevulder thanks! This really a nice cleanup.

Just one comment regarding the iso add registration script. I'd keep it until we do not have a proper alternative and/or until we update our docs.

For all the rest I'd say we can just do the cleanup.

.github/elemental-iso-add-registration
ldevulder: Ok, I will keep it!

As an alternative it's not what `seedImage` is supposed to do? It work, we can customize ISO with it, even in CLI mode.

davidcassany: Yes, but the alternative needs to be a bit more elaborated. If you build your own OS image it is not that straight forward to build ISOs including the registration, as you need to build the ISO without the registration first.

Basically we should review our story around custom images and custom ISOs from custom images.

It is not that it is not possible, just a bit cumbersome or hidden within docs.

davidcassany (approved):
LGTM thanks!

.github/elemental-iso-add-registration
ldevulder: I re-added the scripts, should be fine now.

juadk (approved):

PR #1044, opened 2023-10-15
ci: fix destroy_runner default value

ldevulder:
In scheduled CI `true` needs to be set as a default value.


ldevulder:
Merge forced to fix CI,

Issue #1008, opened 2023-09-19
Move from elemental-teal to slemicro-rancher (based on SLE Micro 5.5)


davidcassany:
This card is moving from elemental-teal to slemicro-rancher images. This implies a change in URLs.

Form:

`registry.opensuse.org/isv/rancher/elemental/dev/containers/rancher/elemental-teal/5.4:latest`

to:

`registry.opensuse.org/isv/rancher/elemental/dev/containers/suse/slemicro-rancher/5.5:latest`

And an equivalent change for the ISO.

Path to do it:

- [x] bump sle micro version in `containers`,  `standard` and ` charts` repos.
- [x] create the new channel to swap to the new image (requires #1002, so ISO URL is no longer needed)
- [x] update and run CI using the new channel
- [x] drop elemental-teal dockerfiles (and associated packages in OBS)
- [ ] tag new development version `2.0-dev`

kkaempf:
That also means we drop `node-image` from Dev/containers and promote `SLE-Micro-Rancher` from Dev/containers_next to Dev/containers !?

kkaempf:
Thinking about it, we probably cannot 'move' but need to 'add'.

Given that we're supposed to adapt to the SLE Micro lifecycle, we would need to maintain "Elemental Teal" (based on SLE Micro for Rancher 5.4) and "SLE Micro for Rancher 5.5" in parallel.

davidcassany:
Indeed, in terms of channels I have been thinking that probably we simply need to add a new channel for the sle micro for rancher and leave teal channel untouched and we simply do not release anything in there anymore (unless required for some maintenance issue)

davidcassany:
Submissions of the new channels
https://build.opensuse.org/request/show/1118217
https://build.opensuse.org/request/show/1118216

davidcassany:
Consider this card completed, no new tag has been added as we probably should discuss new versioning/naming first.

PR #1046, opened 2023-10-17
Few renfinements to the slem4r-os image

davidcassany:
This commit includes 3 changes:
* Move LABELS for the final image instead of intermediate stage
* Sets tags with the usuall 3 tags approach: latest, version and version-release
* Works around ca-certificates installation failures when installing on a custom root

Part of #1008


frelon (approved):

kkaempf (approved):
:+1:

PR #1050, opened 2023-10-17
Introduce "base" variant of SLE-Micro-for-Rancher

kkaempf:
This defines the "base" packages required to run Kubernetes (k3s or rke2) with a minimal number of packages.

Esp. missing is "kernel-firmware-all" which drags in 350MB in packages.

All other SLE Micro for Rancher variants should be derived from "base" and
serve as examples of how to build derived images.

### DIFF
--- a/.obs/dockerfile/slem4r-base-iso/Dockerfile
+++ b/.obs/dockerfile/slem4r-base-iso/Dockerfile
@@ -1,0 +1,44 @@
+# SPDX-License-Identifier: Apache-2.0
+#!BuildTag: suse/sle-micro-rancher-iso/base-%%SLEMICRO_VERSION%%:latest
+#!BuildConstraint: hardware:disk:size unit=G 10
+#!BuildConstraint: hardware:memory:size unit=G 16
+
+ARG SLEMICRO_VERSION
+ARG SLE_VERSION
+
+FROM suse/sle-micro-rancher/base-${SLEMICRO_VERSION}:latest AS os
+FROM suse/sle-micro-rancher/${SLEMICRO_VERSION}:latest AS builder
+
+WORKDIR /iso
+
+COPY manifest.yaml manifest.yaml
+COPY --from=os / rootfs
+
+# Version value is taken form the elemental repository tags
+# Release value of this image and os image are unrelated
+RUN elemental --debug --config-dir . build-iso -o /output -n "sle-micro-rancher.$(uname -m)-base-%VERSION%-Build%RELEASE%" dir:rootfs
+
+# Only keep the ISO as a result
+FROM bci/bci-busybox:$SLE_VERSION
+COPY --from=builder /output /elemental-iso
+
+ARG SLEMICRO_VERSION
+ARG BUILD_REPO=%%IMG_REPO%%
+ARG IMAGE_REPO=$BUILD_REPO/suse/sle-micro-rancher-iso/base-$SLEMICRO_VERSION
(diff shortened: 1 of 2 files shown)


.obs/dockerfile/slem4r-base-os/Dockerfile
```diff
@@ -0,0 +1,104 @@
...
+RUN zypper --installroot /osimage in --no-recommends -y SLE-Micro-Rancher-release systemd-presets-branding-SLE-Micro-for-Rancher
+
+# make elemental-register happy
+RUN zypper --installroot /osimage in --no-recommends -y dmidecode lvm2
+
+# make Rancher (containerd) happy
+RUN zypper --installroot /osimage in --no-recommends -y apparmor-parser
+
+# add elemental
+RUN zypper --installroot /osimage in --no-recommends -y elemental
```
davidcassany, line 43: JFYI here we explicitly require `elemental` package and hence OBS includes this package as part of the buildenv. Then there is the replace_using_package_version OBS buildtime service that finds this package, gets the version of it and replaces a given placeholder with the computed version.

Just something to keep in mind for our own derivatives, we used to compute the tag from `elemental` package version. We could also compute that version|tag from github tags if the sources are pulled from github. This mostly relates to the images we want to build in OBS.

.obs/dockerfile/slem4r-base-iso/Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+FROM suse/sle-micro-rancher/${SLEMICRO_VERSION}:latest AS builder
+
+WORKDIR /iso
+
+COPY manifest.yaml manifest.yaml
+COPY --from=os / rootfs
+
+# Version value is taken form the elemental repository tags
+# Release value of this image and os image are unrelated
+RUN elemental --debug --config-dir . build-iso -o /output -n "sle-micro-rancher.$(uname -m)-base-%VERSION%-Build%RELEASE%" dir:rootfs
```
davidcassany, line 19: In former iso images in OBS this %VERSION% is a placeholder to be replaced with the version computed from github sources, more specific form the `elemental.obsinfo` file. For that the `_service` file requires an obs_scm service and replace_with_package_version service. Something like:

```xml
  <service name="obs_scm">
    <param name="url">https://github.com/rancher/elemental</param>
    <param name="scm">git</param>
    <param name="include">.obs</param>
    <param name="extract">.obs/dockerfile/slem4r-base-iso/Dockerfile</param>
    <param name="revision">main</param>
    <param name="filename">elemental</param>
    <param name="versionformat">@PARENT_TAG@+git%cd.%h</param>
    <param name="versionrewrite-pattern">v([^-]+)-?.*(\+git.*)</param>
    <param name="versionrewrite-replacement">\1\2</param>
  </service>
  <service name="replace_using_package_version" mode="buildtime">
    <param name="file">Dockerfile</param>
    <param name="regex">%VERSION%</param>
    <param name="package">elemental</param>
    <param name="parse-version">patch</param>
  </service>
  ```

davidcassany (approved):
LGTM

.obs/dockerfile/slem4r-base-iso/Dockerfile
kkaempf: Yes, this will be part of the package submission in OBS

kkaempf:
Merging as-is, it doesn't influence any other builds.

PR #1054, opened 2023-10-23
ci/workflow: fix wrong variable type

ldevulder:
`sequential` should be of type boolean instead of string.


davidcassany (approved):
LGTM 🤠

PR #1060, opened 2023-10-24
Multi-flavored OS build

kkaempf:
This builds upon #1050 (the 'base' image) and introduces other flavors
- kvm -- for virtual machines; replaces kernel-default with kernel-default-base, saving ~120 MB
<s>- rt -- for realtime; replaces kernel-default with kernel-default-rt (x86_64 only)</s>
- '' -- the empty flavor; same as 'Elemental Teal' before (aka 'baremetal' with ~350MB of firmware packages

(there's no rt kernel for 5.5)


frelon (approved):
Awesome! Looking forward to the KVM flavor! :tada:

PR #1064, opened 2023-10-26
Update OBS worflow to SLEM4R images

davidcassany:


davidcassany:
looks like the changes in the workflow are not applied within the automated PR build, still using the workflow of main to build this PR 😕 I bet there is reason for such a behavior

frelon (approved):

PR #1071, opened 2023-10-27
Fix SLEM4R OS build

ldevulder:
There is still some minor issues in the build process, SLE Micro version was missing.


davidcassany (approved):
LGTM, thanks!

PR #1078, opened 2023-11-03
Remove longhorn overlay mount

frelon:
As I understand it this path should be mounted by longhorn and we don't need to add a persistent overlay for it.


kkaempf:
This needs to be synced with [longhorn](https://github.com/longhorn/longhorn)

kkaempf (approved):
ack

PR #1090, opened 2023-11-22
Drop 'for Rancher' from image names and tags

kkaempf:
but keep it for the release and systemd-presets package


kkaempf:
not sure if this is acceptable as is - it changes pathes and needs respective `_service` files in OBS adapted.

kkaempf:
Will redo

PR #1091, opened 2023-11-22
Drop the 'for Rancher' suffix

kkaempf:
only for image names and tags.

keep it for the release and the systemd-preset package.

### DIFF
--- a/.obs/dockerfile/slem-flavored-os/Dockerfile
+++ b/.obs/dockerfile/slem-flavored-os/Dockerfile
@@ -1,0 +1,64 @@
+# SPDX-License-Identifier: Apache-2.0
+# Define the names/tags of the container
+#!BuildTag: suse/sle-micro/%%BUILD_FLAVOR%%%%SLEMICRO_VERSION%%:latest
+#!BuildTag: suse/sle-micro/%%BUILD_FLAVOR%%%%SLEMICRO_VERSION%%:%VERSION%
+#!BuildTag: suse/sle-micro/%%BUILD_FLAVOR%%%%SLEMICRO_VERSION%%:%VERSION%-%RELEASE%
+#!BuildConstraint: hardware:disk:size unit=G 8
+#
+
+ARG SLEMICRO_VERSION
+
+FROM suse/sle-micro/base-${SLEMICRO_VERSION}:latest
+
+# dummy zypper call to get elemental into the build context and extract %VERSION% from it via _service
+RUN zypper in --no-recommends -y systemd-presets-branding-SLE-Micro-for-Rancher elemental
+
+# Define labels according to https://en.opensuse.org/Building_derived_containers
+# labelprefix=com.suse.sle.micro
+LABEL org.opencontainers.image.version="%%BUILD_FLAVOR%%%%SLEMICRO_VERSION%%.%VERSION%.%RELEASE%"
+LABEL org.opencontainers.image.created="%BUILDTIME%"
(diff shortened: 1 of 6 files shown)


.obs/dockerfile/slem-os/Dockerfile
```diff
@@ -0,0 +1,102 @@
...
+
+# make dracut happy
+RUN zypper --installroot /osimage in --no-recommends -y squashfs NetworkManager device-mapper iproute2 tar curl ca-certificates ca-certificates-mozilla
+
+# make ARM happy
+#!ArchExclusiveLine: aarch64
+RUN if [ `uname -m` = "aarch64" ]; then zypper --installroot /osimage in -y grub2-arm64-efi raspberrypi-firmware raspberrypi-firmware-config raspberrypi-firmware-dt u-boot-rpiarm64; fi
+
+# make SUSE happy
+RUN zypper --installroot /osimage in --no-recommends -y SLE-Micro-Rancher-release systemd-presets-branding-SLE-Micro-for-Rancher
```
davidcassany, line 34: I guess this should be `SLE-Micro-release` instead of `SLE-Micro-Rancher-release`, could be managed in follow up PRs though.

.obs/dockerfile/slem-iso/manifest.yaml
```diff
@@ -0,0 +1,6 @@
+iso:
+  bootloader-in-rootfs: true
+  grub-entry-name: "SLE Micro for Rancher Install"
```
davidcassany, line 3: I wonder if this should be changed to simply `SLE Micro Install`, in any case I prefer `SLE Micro for Rancher Install` as it clearly points to be used together with Rancher. We could understand `SLE Micro for Rancher` as a use case description rather than product name I guess.

.obs/dockerfile/slem-flavored-os/Dockerfile
```diff
@@ -0,0 +1,64 @@
...
+#!BuildTag: suse/sle-micro/%%BUILD_FLAVOR%%%%SLEMICRO_VERSION%%:%VERSION%-%RELEASE%
+#!BuildConstraint: hardware:disk:size unit=G 8
+#
+
+ARG SLEMICRO_VERSION
+
+FROM suse/sle-micro/base-${SLEMICRO_VERSION}:latest
+
+# dummy zypper call to get elemental into the build context and extract %VERSION% from it via _service
+RUN zypper in --no-recommends -y systemd-presets-branding-SLE-Micro-for-Rancher elemental
```
davidcassany, line 14: Same as in other places, I guess there will be no more `systemd-presets-branding-SLE-Micro-for-Rancher`

davidcassany (approved):
LGTM, just some comments I expect to address in the future.

.obs/dockerfile/slem-os/Dockerfile
kkaempf: Actually no since we want to keep the "for Rancher" suffix in the release package

.obs/dockerfile/slem-iso/manifest.yaml
kkaempf: 🤔

.obs/dockerfile/slem-flavored-os/Dockerfile
kkaempf: I'd keep it for the 5.5 release. To be dropped with 6.0

PR #1095, opened 2023-11-27
Rename build-iso-next

davidcassany:
Renaming build-iso-next to build-iso as this is not any longer the 'next' version of the ISO. Just to prevent confusions.


davidcassany:
The `build-iso-next` package is already copied to a new package called `build-iso`. Once merged I will just delete the `build-iso-next` package from Dev.

frelon (approved):

PR #1096, opened 2023-11-27
Remove obsolete images, sle micro variants are already in place

davidcassany:


davidcassany:
Removing `slem4r` images, the `slem` only variants are already in place.

kkaempf (approved):
Yeah, kill it with 🔥

PR #1098, opened 2023-11-27
Include real time version as flavor

davidcassany:
With this change rt versions of the image will referenced as an additional flavor:
 `suse/sle-micro/rt-5.5:<version>`
 
 This also allows us to drop the RT package in OBS as it will be just a flavor of SLE-Micro image.

### DIFF
--- a/.obs/dockerfile/slem-flavored-os/Dockerfile
+++ b/.obs/dockerfile/slem-flavored-os/Dockerfile
@@ -17,6 +17,8 @@ RUN zypper in --no-recommends -y systemd-presets-branding-SLE-Micro-for-Rancher 
 RUN if [[ "%%BUILD_FLAVOR%%" == "kvm-" ]]; then \
   # replace default kernel with smaller one, sufficient to run kvm guests \
   zypper in --no-recommends -y kernel-default-base -kernel-default; \
+elif [[ "%%BUILD_FLAVOR%%" == "rt-" ]]; then \
+  zypper in --no-recommends -y kernel-rt -kernel-default; \ 
 else \
   # extend -base with baremetal and usability packages \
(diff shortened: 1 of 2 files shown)


.obs/dockerfile/slem-flavored-os/Dockerfile
```diff
@@ -17,6 +17,9 @@ RUN zypper in --no-recommends -y systemd-presets-branding-SLE-Micro-for-Rancher
 RUN if [[ "%%BUILD_FLAVOR%%" == "kvm-" ]]; then \
   # replace default kernel with smaller one, sufficient to run kvm guests \
   zypper in --no-recommends -y kernel-default-base -kernel-default; \
+elif [[ "%%BUILD_FLAVOR%%" == "rt-" ]] && [ $(uname -m) = "x86_64" ]; then \
+  #!ArchExclusiveLine x86_64
```
davidcassany, line 21: I can't manage to make this like functional together with flavors... 🤷🏽‍♂️ I did many different attempts without any luck

kkaempf: Welcome to the club (of people failing to get build flavors to work in Dockerfiles)

davidcassany: Yes... so far I found two different issues... First, I can't make a per-arch _multibuild (or at least I could not find it in OBS docs) to enable the `rt-` only for x86_64. Second, for some reason the `#!ArchExclusiveLine x86_64` doesn't seam to be functional if used together with flavors, this is really odd as I don't see the relation 🤷🏽‍♂️

davidcassany:
Closing, I give up for now

PR #1108, opened 2023-11-28
For some reason OBS complains of double quotes on SOURCEURL substitution

davidcassany:


frelon (approved):

davidcassany:
A portion of the logs if using the following line (without quotes)

```Dockerfile
LABEL org.opencontainers.image.source=%SOURCEURL%
```

build logs:
```
[   97s] Step 32/43 : LABEL com.rancher.slem.org/package/show/isv:Rancher:Elemental:PR:rancher:elemental:PR-1108/build-iso-base?rev=d972e5040b2c5ebeda2c0ee4234c6eae
[   97s]  ---> Running in ad64c13f98e5
[   97s] Removing intermediate container ad64c13f98e5
[   97s]  ---> ae65dbbf86d8
[   97s] Step 33/43 : LABEL org.opencontainers.image.source=https://build.opensuse.org/package/show/isv:Rancher:Elemental:PR:rancher:elemental:PR-1108/build-iso-base?rev=d972e5040b2c5ebeda2c0ee4234c6eae
[   97s]  ---> Running in a4ec953ca9dc
```

See there is a wrong substitution... that's why if we use quotes it doesn't work as first quotes are swallowed by the substitution and it results into an invalid format. Removing the quotes makes the build to succeed but then we get this crazy labels... IMHO better removing the source label until we have a proper fix (which is likely to point a bug in labeling OBS service).

PR #1117, opened 2023-12-05
Remove default serial console kernel parameter

davidcassany:


davidcassany:
@kkaempf I do believe that to keep serial console in RPi we should handle that in a separate PR and eventually use that also as documentation exercise. So we document how to include extra parameters in the kernel command line.

kkaempf (approved):
👍🏻

PR #1124, opened 2023-12-11
ci/cli: fix OBS Staging tests

ldevulder:
`Staging` is a mix between `Dev` and `Stable` so it needs some specifics configuration to work.

Verification run:
- [CLI-K3s-OBS_Staging](https://github.com/rancher/elemental/actions/runs/7169104514) :white_check_mark:
- [CLI-K3s-OBS_Staging (hardened)](https://github.com/rancher/elemental/actions/runs/7171033111) :white_check_mark:
- [CLI-RKE2-OBS_Staging](https://github.com/rancher/elemental/actions/runs/7169419838) :white_check_mark:
- [CLI-RKE2-OBS_Staging (hardened](https://github.com/rancher/elemental/actions/runs/7170729300) :white_check_mark:
- [CLI-K3s-OS-Upgrade (with upgrade to Rancher Manager 2.7-HEAD)](https://github.com/rancher/elemental/actions/runs/7169090069) :negative_squared_cross_mark: Fails because of [#577](https://github.com/rancher/elemental-operator/issues/577)
- [CLI-RKE2-OS-Upgrade (with upgrade to Rancher Manager 2.8-HEAD)](https://github.com/rancher/elemental/actions/runs/7169612255) :negative_squared_cross_mark: Fails because of [#577](https://github.com/rancher/elemental-operator/issues/577)

### DIFF
--- a/.github/workflows/cli-k3s-obs_staging.yaml
+++ b/.github/workflows/cli-k3s-obs_staging.yaml
@@ -40,5 +40,5 @@ jobs:
       destroy_runner: ${{ inputs.destroy_runner }}
       k8s_version_to_provision: v1.26.10+k3s2
       operator_repo: oci://registry.opensuse.org/isv/rancher/elemental/staging/charts/rancher
-      os_to_test: staging
+      os_to_test: unstable
       rancher_version: ${{ inputs.rancher_version }}
--- a/.github/workflows/cli-k3s-os-upgrade.yaml
+++ b/.github/workflows/cli-k3s-os-upgrade.yaml
@@ -14,7 +14,7 @@ on:
         type: boolean
       operator_repo:
         description: Operator version to use for initial deployment
-        default: oci://registry.opensuse.org/rancher
+        default: oci://registry.suse.com/rancher
         type: string
       rancher_upgrade:
         description: Rancher Manager channel/version to upgrade to
@@ -58,3 +58,4 @@ jobs:
       rancher_upgrade: ${{ inputs.rancher_upgrade }}
       rancher_version: ${{ inputs.rancher_version }}
       upgrade_image: registry.opensuse.org/isv/rancher/elemental/${{ inputs.upgrade_os_channel }}/containers/suse/sle-micro/${{ inputs.slem_version }}:latest
+      upgrade_os_channel: ${{ inputs.upgrade_os_channel }}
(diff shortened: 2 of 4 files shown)


.github/workflows/cli-k3s-obs_staging.yaml
```diff
@@ -40,5 +40,5 @@ jobs:
       destroy_runner: ${{ inputs.destroy_runner }}
       k8s_version_to_provision: v1.26.10+k3s2
       operator_repo: oci://registry.opensuse.org/isv/rancher/elemental/staging/charts/rancher
-      os_to_test: staging
+      os_to_test: unstable
```
davidcassany, line 43: I am wondering if this has impact on the description of this value on other workflows

```
$ grep -rn "OS repository to test (dev/staging/stable)" .github/
.github/workflows/cli-obs-manual-workflow.yaml:31:        description: OS repository to test (dev/staging/stable)
.github/workflows/master-e2e.yaml:90:        description: OS repository to test (dev/staging/stable)
```
Which are the valid values of `os_to_test`? How is this used within the tests? is this just used in a regexp over the available OSVersions? Probably the description include the regexp instead of a list of values that can be outdated.

davidcassany (approved):
Looks good to me

Just a doubt about the `os_to_test` value, probably we should review the description of this value on some workflows.

.github/workflows/cli-k3s-obs_staging.yaml
ldevulder: At the beginning it was the value to set when all was on OBS (Dev/Staging/Stable) and with a `latest` link to latest OS image/OS. But is now not the case anymore, mainly because we don't have any `latest` tag/label and `unstable` label is in fact set in `Dev` and `Staging` but not in `Stable`.

So when `Stable` is set the latest image/OS version is used, when `Dev` is set the `unstable` label is searched and used if found and the same for `Staging`. But `dev` is more an alias to `unstable`  for the OS image name.

This variable can be used in installation and upgrade tests, but can act differently depending of what we have to test (keep in mind that an upgrade test will *always* install `Stable` image/OS first and then upgrade to `Dev/Staging`).

Don't know if I'm clear enough ;-) But the use of this variable has been changed upon time mainly because of the different ways to handle `Dev/Staging/Stable` with OBS/IBS. For sure at some point we will have to make it clearer I think, but we didn't had enough time. With the changes that CAPI will imply we already thinking of rewriting most of the code and so trying to get rid of all that mess ;-)

PR #1125, opened 2023-12-12
Add default network configuration for eth0

davidcassany:
This adds network the configuration file that got removed from the `cloud-config-essentials` feature in elemental-toolkit in rancher/elemental-toolkit@1a1b942acdc05d5351d89a38610c44016d1af38b

### DIFF
--- a/framework/files/system/oem/05_network.yaml
+++ b/framework/files/system/oem/05_network.yaml
@@ -1,0 +1,12 @@
+name: "Default network configuration"
+stages:
+   initramfs:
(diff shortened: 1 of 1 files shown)


davidcassany:
IMHO this sort of configurations should not be part of the `elemental init` they are pretty use case specific. Probably we should also consider other strategies of the default network configuration so we do not assume the existence of `eth0`.

frelon (approved):

ldevulder (approved):

PR #1131, opened 2023-12-15
Revert to 'old' RPi firmware

kkaempf:
ARM's 'system ready' firmware is not ready for prime time yet:

It also only supports ACPI at this time, but not device tree.


davidcassany (approved):
LGTM

PR #1145, opened 2023-12-22
Run services after time-sync.target

frelon:
Change elemental-register-install and elemental-register-reset to run after systemd time-sync.target.

This is needed to have the correct time before trying to register using https certificates that are time sensitive.

We also enable the systemd-time-wait-sync service to wait until time is synced before time-sync.target is reached.


anmazzotti (approved):
Nice one!

Issue #1138, opened 2023-12-21
"wicked" network service starting


kkaempf:
There's still `wicked` being started at bootup. This needs to go, only NetworkManager is supported.

kkaempf:
`wicked` and `wicked-service` are installed, pulled in by some dependency 🙈 

Will fix it together with #1141 🤞🏻

kkaempf:
It's a new dependency of `dracut` 🤦🏻 
~~~
> rpm -q --requires dracut | grep wic
(iputils or arping2 or wicked)
~~~

so adding `iputils` should fix that ...

PR #1147, opened 2023-12-22
Add openssh-server

### LINKED ISSUE #1141: Default image for 5.5 is missing openssh-server
The default image (i.e. the one without any suffix like `-base`, `-rt`, or `-kvm`) should have the same package set (and the same enabled services) as SLE Micro for Rancher 5.4 had.
While not ideal, we want to keep backward compatible while on 5.x 😉

kkaempf:
Fixed #1141

(also drop iputils because [it's included in base](#1138))


frelon (approved):
LGTM

PR #1171, opened 2024-01-19
Don't set the hostname

### LINKED ISSUE #1172: Don't set the hostname
In the Elemental framwork we set the hostname to rancher-${RANDOM}.
While this is fancy as the user will see on the booted machine console a properly set hostname (which otherwise may be set to "localhost"), these prevents DHCP to serve a transient hostname to the host.
In Elemental, during the registration phase, the hostname is recorded as the MachineInventory name and from that point on it could not be changed. It will be later enforced as the static hostname during the k3s/RKE2 provisioning.

Since we do registration just after the ISO boot, in order to allow usage of a DHCP provided hostname, we should avoid to provide a static hostname so that the DHCP client will set the (transient) hostname: at that point, one can take advantage of the new templating value: ${System Data/Runtime/Hostname} introduced in  https://github.com/rancher/elemental-operator/pull/593 to set the hostname (during kube provisioning) to the DHCP provided value.

Drawbacks: if no hostname is provided by the DHCP server or retrieved via IP reverse lookup (NetworkManager tries this too), if the MachineInventory MachineName is not set, the Elemental host will have "localhost" hostname till k3s/RKE2 provisioning is started (right now it stays "rancher-${RANDOM}". During k3s/RKE2 provisioning, with no MachineName provided, the usual m-${UUID} hostname is set.

PR bringing the required change (dropping the snippet setting the hostname) here #1171 .
Let's discuss here if we want to drop the default rancher-${RANDOM} hostname.

fgiudici:
If the (static) hostname is not set, we used to set a rancher-${RANDOM} one. This anyway prevents DHCP to serve a transient hostname. In Elemental, at registration time, the hostname is recorded as the MachineInventory name and from that point onwards it could not be changed. It will be enforced as the static hostname during the k3s/RKE2 provisioning.
Since we do registration just after the ISO boot, in order to use the DHCP provided hostname, we should allow the transient name to be set by the dhcp client (which happens only if the static hostname is not set) and then take advantage of the new templating value: ${System Data/Runtime/Hostname}
(see https://github.com/rancher/elemental-operator/pull/593)

Let's anyway set the racher-${RANDOM} hostname but as a transient one  and before the network stage (so it may be replaced by the DHCP provided one).

This is a change in behavior: on Elemental booting (before provisioning) if a hostname has not been set and DHCP provides one, the DHCP hostname takes precedence over the rancher-${RANDOM} one.

Fixes #1172

### DIFF
--- a/framework/files/system/oem/01_elemental-rootfs.yaml
+++ b/framework/files/system/oem/01_elemental-rootfs.yaml
@@ -5,8 +5,7 @@ stages:
       commands:
       - |
         if [ ! -e /run/elemental/persistent/etc/hostname ]; then
-          mkdir -p /run/elemental/persistent/etc
-          echo rancher-${RANDOM} > /run/elemental/persistent/etc/hostname
+          sysctl kernel.hostname=rancher-${RANDOM}
         fi
         ln -sf /run/elemental/persistent/etc/hostname /etc/hostname
     - if: '[ ! -f "/run/elemental/recovery_mode" ]'


frelon (approved):
LGTM

fgiudici:
Patch changed: we keep creating the symlink and we set a transient hostname.

frelon (approved):

PR #1176, opened 2024-01-22
ci/qase: some fixes and improvements with qase/ginkgo integration

ldevulder:
It is not very useful to keep Qase run/report when a job/workflow has been canceled, as the run will never be completed and more over we could have all cases green in the report and some missing so the whole report will be green and that's not the real status of the test.
It is also better to publish the whole test result at the end of the test.

List of all fixes:
- qase/lib: fix some checks
- ci: add GH test run URL in Qase result
- ci: publish and complete Qase Run outside of Ginkgo
- ci: delete Qase Run if job has been cancelled
- qase/lib: update README
- qase/lib: remove QASE_DELETE_RUN support

Verification runs:
- [CLI-K3s-OBS_Dev (cancelled, so Qase result should be deleted](https://github.com/rancher/elemental/actions/runs/7624616258)
- [CLI-RKE2-OBS_Dev (not cancelled, so Qase result should be available)](https://github.com/rancher/elemental/actions/runs/7625531659)

**NOTE:** 2nd VR failed for another issue, not related to this PR.


tests/qase/qase_cmd.go
```diff
@@ -0,0 +1,68 @@
...
+		logrus.Debugf("Qase run id %d created", id)
+		fmt.Printf("%d", id)
+	} else if *deleteRun {
+		qase.DeleteRun()
+		logrus.Debugf("Qase run id %d deleted", runID)
+	} else if *publishRun {
+		qase.FinalizeResults()
+		logrus.Debugf("Qase finalization for run id %d has been done", runID)
+	} else {
+		logrus.Debugln("Nothing do to!")
```
juadk:
```suggestion
		logrus.Debugln("Nothing to do!")
```

juadk (approved):

tests/qase/qase_cmd.go
ldevulder: Fixed, thanks!

PR #1177, opened 2024-01-23
Update rootfs settings

frelon:
Moving persistent mount to /run/elemental/persistent makes /usr/local readonly. This commit makes /usr/local a writeable persistent mount.


kkaempf (approved):
:+1:

PR #1178, opened 2024-01-24
Make initrd generation consistent across flavors

davidcassany:


frelon (approved):

davidcassany:
Tested a manual run and installation of the resulting images from the builds at https://build.opensuse.org/project/show/isv:Rancher:Elemental:PR:rancher:elemental:PR-1178 and at least installs and the installed system boot.

Not sure if this alone will mostly fix Dev env, but this is certainly an improvement.

PR #1191, opened 2024-02-02
Add back recovery hostname

### LINKED ISSUE #608: When booting recovery, prompt does not show recovery anymore
With the latest dev operator, in recovery mode, the prompt is not `recovery login` anymore, it's `localhost login`.

With staging:
```
ssh root@192.168.1.116
(root@192.168.1.116) Password:
recovery:~ #
```

frelon:
Hostname is no longer set since the branding files in the toolkit was moved to `cloud-config-defaults` feature which we no longer install in elemental.

Fixes https://github.com/rancher/elemental-operator/issues/608

### DIFF
--- a/framework/files/system/oem/03_branding.yaml
+++ b/framework/files/system/oem/03_branding.yaml
@@ -1,0 +1,7 @@
+name: "Branding"
+stages:
+   boot:
+    - name: "Recovery"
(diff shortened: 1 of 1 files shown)


ldevulder (approved):
Thanks!

PR #1194, opened 2024-02-05
ci/cli: fix regression on operator installation

ldevulder:
Fix regression added in #1189. But weirdly we have no sign of Elemental Operator installation even before this PR, it seems to have disappear since the Qase/Ginkgo integration... But the Operator was installed, otherwise tests would have failed before.

Anyway, it seems that a `if` check in Ginkgo test before a `It` is not working, even if we have no syntax error!

Verification runs:
- [CLI-K3s-RM_Stable](https://github.com/rancher/elemental/actions/runs/7781871562)
- [UI-K3s-RM_Stable](https://github.com/rancher/elemental/actions/runs/7782492119)

### DIFF
--- a/tests/e2e/install_test.go
+++ b/tests/e2e/install_test.go
@@ -225,7 +225,7 @@ var _ = Describe("E2E - Install Rancher Manager", Label("install"), func() {
 		}
 	})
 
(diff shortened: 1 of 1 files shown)


juadk (approved):
LGTM thx!

PR #1195, opened 2024-02-05
ci/ui: increase timeout for reset test

juadk:
The reset test fails sporadically when waiting for the new machine inventory to be active.
This PR increases the timeout a bit.


juadk:
Finally, it is not a timeout issue => https://github.com/rancher/elemental/issues/1210

PR #1213, opened 2024-02-14
Adapt bootargs file to support btrfs snapshotter

davidcassany:


kkaempf (approved):
lgtm

davidcassany:
Requires https://github.com/rancher/elemental-toolkit/pull/1957 to merged first/together

frelon (approved):

.obs/dockerfile/slem-base-os/Dockerfile
```diff
@@ -44,6 +44,9 @@ RUN zypper --installroot /osimage in --no-recommends -y apparmor-parser
 # add elemental
 RUN zypper --installroot /osimage in --no-recommends -y elemental
 
+# add transactional updates utilities
+RUN zypper --installroot /osimage in --no-recommends -y transactional-update
```
davidcassany: Transactional update stack added here to be capable to explore the possibility of running elemental upgrade **and** transactional updates over the same system without breaking the system. 

IMHO, this is a nice to have not mandatory urgent goal. Adding it here allows us to verify installing it does not break anything and it also allows to easily manually test it afterwards. I don't think it will work out of the box as the mount layout between SLE Micro and Elemental is slightly different (slightly different assumptions on `/var` and on `/etc`).

davidcassany: Reverted this change as I could actually validate transactional-update is not functional. Snapper is, but not transactional-update, libtukit fails as it has some expectations on the underlaying layout which are not met (mostly around /var and /etc paths).

PR #1228, opened 2024-02-21
ci: fix manual OBS workflow

ldevulder:
This commit will use the default version, which is K3s and most of the tests are done with this distribution for now.


davidcassany (approved):
LGTM

PR #1229, opened 2024-02-21
ci: use new kubectl.RunWithoutErr function

ldevulder:
As we already rely on the return code to trap the failure, and the current implementation can lead to sporadic issues in some cases, it is better to simply use the new RunWithoutErr function and get rid of the stderr output.

Verification runs:
- [CLI-K3s-RM_head_2.7](https://github.com/rancher/elemental/actions/runs/7989240495)
- [UI-RKE2-RM_Stable](https://github.com/rancher/elemental/actions/runs/7987764209)


fgiudici (approved):
LGTM

PR #1232, opened 2024-02-22
Include btrfsprogs in dev image

anmazzotti:
Without btrfsprogs `mkfs.btrfs` is not found and we can't install. 
This occurs with dev `elemental-toolkit` since it defaults to btrfs.


davidcassany (approved):
LGTM

PR #1236, opened 2024-02-22
Set transient recovery hostname

### LINKED ISSUE #1242: Booting into recovery sets static hostname
When resetting an Elemental machine provisioned using the new machineName key '${System Data/Runtime/Hostname}' the hostname used is "recovery".

The system to use the DHCP hostname.

Introduced here: https://github.com/rancher/elemental/pull/1191

**Environment:**

- Elemental release version (use `cat /etc/os-release`): Dev
- Rancher version: 2.8.2
- Kubernetes version (use `kubectl version`): N/A
- Cloud provider or hardware configuration: QEMU/KVM

frelon:
Only setting hostname for recovery using yip overwrites the hostname and uses "recovery" also for active mode.

This commits sets a transient hostname when booting into recovery, hopefully allowing active hostname to not be overwritten.

### DIFF
--- a/framework/files/system/oem/03_branding.yaml
+++ b/framework/files/system/oem/03_branding.yaml
@@ -3,5 +3,6 @@ stages:
    boot:
     - name: "Recovery"
       if: '[ -f "/run/elemental/recovery_mode" ]'
-      hostname: "recovery"
+      commands:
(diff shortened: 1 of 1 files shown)


fgiudici (approved):
Looks good to me!

PR #1264, opened 2024-03-04
Always ensure a Rancher-vWXYZ transient hostname

### LINKED ISSUE #646: recovery as machine inventory name after reset
I was playing with my rancher / elemental lab using` machineName: ${System Data/Runtime/Hostname}` and when I reset the node, I saw that machine inventory got created with recovery name (see screenshot below)

image

What will happen if we reset multiple nodes at the same time? There will likely be conflict because we cannot have multi-machine inventory with the same name... 

It might be fixed by https://github.com/rancher/elemental/pull/1249 but we will have to confirm it.

fgiudici:
If a static hostname is not available always set a transient hostname in the form "Rancher-VWXYZ". Do that also when in recovery mode.

Fixes: https://github.com/rancher/elemental-operator/issues/646

### DIFF
--- a/framework/files/system/oem/01_elemental-rootfs.yaml
+++ b/framework/files/system/oem/01_elemental-rootfs.yaml
@@ -3,11 +3,7 @@ stages:
   initramfs:
     - if: '[ ! -f /run/elemental/recovery_mode ]'
       commands:
-      - |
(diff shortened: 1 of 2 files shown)


frelon (approved):
Very nice!

PR #1271, opened 2024-03-06
Actualize development documentation

anmazzotti:
Added a separate dev machineregistration to have both snapshotters ready to be tested out of the box.

Also added some tips and tricks to verify the upgrade process.


davidcassany (approved):
Nice!

PR #1275, opened 2024-03-08
Include busybox to ISO images from scratch

davidcassany:


davidcassany:
I think this does not create a relevant maintenance overhead on our side as we are still in a single image approach and we shift the busybox image requirement to busybox package requirement which is easier to get I guess.

kkaempf (approved):
😮

PR #1277, opened 2024-03-11
ci: allow root login by SSH

davidcassany:


kkaempf (approved):
Should be added to docs as well, showing our users how to enable root logins in Micro 6.

davidcassany:
Noticed that by default on latest SSH root is not allowed to login, requires explicit configuration. This prevents the CI to get stuck attempting to connect to nodes on reboot

davidcassany:
> Should be added to docs as well, showing our users how to enable root logins in Micro 6.

Indeed, that would be nice to be documented, together with adding a serial console at boot. It was quite annoying not having the serial console while debugging. Gonna verify this change first, next is adding a serial console for tests, so we get all boots and reboots data.

PR #1283, opened 2024-03-13
Only apply upgrade config on OS upgrades

anmazzotti:
Part of rancher/elemental-operator#670

Note that this is not strictly necessary. 
Removing this line only serves to not create false expectations on the user, regarding the application of the cloud-init config when the system is not actually upgraded.  

Note that this still applies to the UPGRADE_RECOVERY_ONLY scenario, but in that case the config is already "effective" even if the system will not reboot, since it only applies to recovery boots.

### DIFF
--- a/framework/files/usr/sbin/suc-upgrade
+++ b/framework/files/usr/sbin/suc-upgrade
@@ -72,7 +72,6 @@ function isEqualVersion() {
         echo "Upgrade already done with release:"
         cat ${HOST_DIR}${RELEASE_FILE}
 
-        config
         exit 0
     fi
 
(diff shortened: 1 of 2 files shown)


anmazzotti:
>I'm still not sure whether this is a feature or not.
>
>Someone could theoretically update this config by simply updating the `ManagedOSImage.spec.cloudConfig`. 
>That will make the downstream Plan mutate and be reapplied (with the same osImage, hence equal version). 
>
>The user can take the manual step of rebooting the host, or maybe exploit this in other ways, like having a cron job that does invoke `elemental run-stage boot`. 

Actually scratch that. The Plan will not mutate, just one of the upstream Bundle resources, so editing the cloudConfig only will not have any effect. 

I think that answers my doubts and this can be removed.

davidcassany (approved):
Nice, looks good to me

PR #1291, opened 2024-03-14
Include audit package and boot selinux in permissive mode explicitly

### LINKED ISSUE #1289: Set SELinux in permessive mode
Explicitly set SELinux in permissive mode and try to properly document audit logs so they are not that prominent in journalctl. IIRC this can be done by adding some of the SELinux audit packages, need to be careful we are not pulling the python stack with them.

Probably I'd set it only for the _v2.1.x_ (Staging) series and leave _main_ (Dev) with the OS default

davidcassany:
Fixes #1289


kkaempf (approved):
lgtm

davidcassany:
Verified that now I don't see SELinux denials on journal and stdout, they are stored in audit logs.

PR #1293, opened 2024-03-15
Micro rename

kkaempf:
First step 🤞🏻


davidcassany (approved):
The beauty of renaming.... it is quite hard to verify if there is some leftover somewhere. 

In any case this PR looks good :+1:

PR #1295, opened 2024-03-15
Install openssl in micro-base-os

frelon:
Moves installation of openssl from micro-os to micro-base-os.

This makes the kvm flavor have openssl which is needed for machine registration.


kkaempf (approved):
lgtm

PR #1299, opened 2024-03-18
ci/ui: fix elemental operator upgrade

juadk:
- Fix this failing test https://github.com/rancher/elemental/actions/workflows/ui-k3s-os-upgrade-rm_stable.yaml
It looks like one HTML selector has changed.

- Improves the code to get the latest elemental dev version instead of using a hardcoded version

[UI-K3s-OS-Upgrade-RM_Stable](https://github.com/rancher/elemental/actions/runs/8328310769) ✅

### DIFF
--- a/.github/workflows/sub_ui.yaml
+++ b/.github/workflows/sub_ui.yaml
@@ -163,7 +163,15 @@ jobs:
 
       - name: Install Chartmuseum
         id: install_chartmuseum
-        run: cd tests && make e2e-install-chartmuseum
+        run: |
+          cd tests && make e2e-install-chartmuseum
+          # Extract latest dev operator version
+          OPERATOR_DEV_CHART=$(ls elemental-operator-chart*)
+          i=${OPERATOR_DEV_CHART##*-}
+          OPERATOR_DEV_VERSION=${i%.tgz}
+
+          # Export value
+          echo "operator_dev_version=${OPERATOR_DEV_VERSION}" >> ${GITHUB_OUTPUT}
 
       # Basics means tests without an extra elemental node needed
       - name: Cypress tests - Basics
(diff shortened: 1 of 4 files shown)


.github/workflows/sub_ui.yaml
```diff
@@ -163,7 +163,15 @@ jobs:
 
       - name: Install Chartmuseum
         id: install_chartmuseum
-        run: cd tests && make e2e-install-chartmuseum
+        run: |
+          cd tests && make e2e-install-chartmuseum
+          # Extract latest dev operator version
+          OPERATOR_DEV_CHART=$(ls elemental-operator-chart*)
```
ldevulder, line 169: I would have add a `-` before the `*`, even if I'm not sure that we may have something like:
```bash
$ ls
elemental-operator-chart-1.5.0.tgz elemental-operator-charts-1.6.0.tgz
```

tests/cypress/latest/e2e/unit_tests/upgrade-operator.spec.ts
```diff
@@ -52,14 +52,14 @@ describe('Elemental operator upgrade tests', () => {
           cy.clickButton('Upgrade');
           cy.contains('.header > .title', 'elemental-operator');
           cy.clickButton('Next');
-          cy.get('[data-testid="string-input-channel.repository"]')
+          cy.get('[data-testid="string-input-channel.image"]')
             .clear()
-          cy.get('[data-testid="string-input-channel.repository"]')
-            .type('rancher/elemental-channel')
+          cy.get('[data-testid="string-input-channel.image"]')
+            .type('registry.opensuse.org/isv/rancher/elemental/dev/containers/rancher/elemental-channel')
```
ldevulder, line 58: So not possible to test upgrade from `Stable` to `Staging` with this hardcoded? Maybe something like [here](https://github.com/rancher/elemental/blob/9a5f23fc12822fb581f305ad64d79459c02c0bbc/tests/cypress/latest/support/commands.ts#L448) could be possible?

ldevulder (approved):
LGTM! Just two things maybe for another PR, as this one fixes the CI.

tests/cypress/latest/e2e/unit_tests/upgrade-operator.spec.ts
juadk: Yes, I was aware of this one, I will open an issue to improve it, thx.

PR #1306, opened 2024-03-19
Finalize Micro rename

kkaempf:
- change tags to suse/sl-micro/<micro-version>/<flavor>-os-container resp. suse/sl-micro/<micro-version>/<flavor>-iso-image

- change ISO image filenames to include SL Micro version

- drop the "non flavored" flavor, call it "baremetal"

- adapt grub boot entry names (in manifest.yaml) accordingly

- iron out any inconsistencies


davidcassany (approved):
LGTM

fgiudici (approved):

PR #1310, opened 2024-03-21
Unblocking the build

kkaempf:


kkaempf:
Hmm, looks like autobuild team found a workaround

PR #1315, opened 2024-03-22
Adding an network-online checker for elemental

### LINKED ISSUE #1263: Boot assessment must wait for network
We have a user case where the system comes up, passes boot assessment, but fails to start the workload.

Turns out that there's a filesystem error, preventing even NetworkManager.service from starting.

Boot assessment should capture this case and reboot into `fallback` if network fails to start up.

davidcassany:
With this checker if the network-online.target does not reach the active state the system will reboot to fallback.

Few considerations for current boot_assessment:

1. Passive will also run the boot_assessment on upgrade in case active fails. This is relevant as it allows the system to keep trying on different fallbacks (assuming there are more than one old snapshot). However this also implies that if none is functional it will eventually reach recovery system. Before it used to only stay on a single fallback only, never reaching recovery automatically, because the boot assessment was not executed on passive.

2. Boot_assessment is not active by default on regular boots, just for reboots after upgrading. This is relevant as the boot assessment is not executed after install. So it could happen that the boot assessment is not passing and you just realize this on an upgrade. In that case the system will land in to recovery mode (out of k8s) after upgrading.

So I am wondering if it would make sense defaulting the boot_assessment to always run, not only after upgrades. This way we assess the installed boot actually passess the boot assessment.

@frelon @anmazzotti @fgiudici any thougths about this?

Fixes #1263


fgiudici:
🤔 the scenario that worries me is if there is some transient network connectivity issue (some remote edge deployment with unstable connectivity?).
Will that mean that a system may end up in recovery state in that case?

davidcassany:
> 🤔 the scenario that worries me is if there is some transient network connectivity issue (some remote edge deployment with unstable connectivity?). Will that mean that a system may end up in recovery state in that case?

Good point, that could happen if the network outage happen during the upgrade reboot, which is a corner case as network required to run the upgrade. The window in which this could happen is rather small, but I guess at scale this is not that hard hitting such a corner case in few nodes...

Yes this is all annoying. That's why I am hesitant to include certain checks, it could easily hit us back with undesired effects on certain corner cases. Also it is unclear what `network-online.target` encompasses.

frelon:
We could also use the `boot-complete.target` since that seems to be what the systemd boot assessment [uses](https://systemd.io/AUTOMATIC_BOOT_ASSESSMENT/).

davidcassany:
> We could also use the `boot-complete.target` since that seems to be what the systemd boot assessment [uses](https://systemd.io/AUTOMATIC_BOOT_ASSESSMENT/).

This is quite interesting, but this seams to relay on https://uapi-group.org/specifications/specs/boot_loader_specification which we are not following (if we do it is by pure coincidence). Following this spec would be nice though, I wonder if it is possible to follow it with grub2 and systemd-boot. If so then we could have our bootloader setup agnostic to the underlaying bootloader, which would be pretty nice. I think this require some deeper investigation.

davidcassany:
@fgiudici in https://github.com/rancher/elemental-toolkit/pull/2027 the boot_assessment changes to only iterate over existing snapshots without including recovery. Hence on reboot after install, upgrade or reset it will kick-in the boot assessment and run checks if any. In case of failure it will reboot to most recent fallback snapshot and from there keep trying until there is a successful boot or until the oldest snapshot. In the oldest snapshot if it fails to boot it will essentially keep rebooting every few minutes in an infinite loop from that old snapshot.

Probably instead of being stuck to the oldest we could consider restarting from scratch, active and then falling to most recent non active snapshot up to the oldest again. I'd consider that a minor detail though, if no snapshot is booting there isn't much we can do...

Toughts?

frelon:
> This is quite interesting, but this seams to relay on https://uapi-group.org/specifications/specs/boot_loader_specification which we are not following (if we do it is by pure coincidence). Following this spec would be nice though, I wonder if it is possible to follow it with grub2 and systemd-boot. If so then we could have our bootloader setup agnostic to the underlaying bootloader, which would be pretty nice. I think this require some deeper investigation.

In the [docs](https://www.freedesktop.org/software/systemd/man/latest/systemd.special.html) it's stated that it's a generic synchronization point so it _should_ work using any bootloader. We would probably need to `Require=` it somewhere in our boot to actually activate it.

fgiudici:
> @fgiudici in [rancher/elemental-toolkit#2027](https://github.com/rancher/elemental-toolkit/pull/2027) the boot_assessment changes to only iterate over existing snapshots without including recovery. Hence on reboot after install, upgrade or reset it will kick-in the boot assessment and run checks if any. In case of failure it will reboot to most recent fallback snapshot and from there keep trying until there is a successful boot or until the oldest snapshot. In the oldest snapshot if it fails to boot it will essentially keep rebooting every few minutes in an infinite loop from that old snapshot.
> 
> Probably instead of being stuck to the oldest we could consider restarting from scratch, active and then falling to most recent non active snapshot up to the oldest again. I'd consider that a minor detail though, if no snapshot is booting there isn't much we can do...
> 
> Toughts?

So, it makes sense then, if things go awry we may end-up with an older snapshot, one could always retry an OS upgrade later 👍🏼 
Why the network-online.target then? Is it because it is the last one?
The proposal from @frelon to use the boot-complete.target sounds more "generic" (also if the result would be the same, if I understood it correctly). And sure, we are not going all the systemd schema to manage rollbacks, so no real need to follow that, we would just align to the same target service from systemd.
But anyway, I think rebooting with a previous snapshot will manage the corner case well.
We have to remember to proper document that, in order to make it pretty clear if anyone uses Elemental in special scenarios (edge?).

fgiudici (dismissed):
LGTM! Approving from my side 👍🏼 
Would like anyway to get some more on @frelon PoV.

davidcassany:
> Why the network-online.target then? Is it because it is the last one?

Because we saw a customer with a failed upgrade (filesystem corruption) and it booted with many failed services (dbus, networkmanager, etc), but since none of these is required it just booted in degraded mode and it was not reachable anymore. Boot assessment did not catch the error because the system booted from systemd perspective. Here is where the health check concept comes from. We could also consider having a health check on elemental-register service instead of network-online.

If we stick to elemental-register check that means that a node that is not capable to re-register after (re)booting will try to restart to a fallback snapshot. That could also be a more elemental specific approach.

fgiudici (approved):
Well, the elemental-register sync-up after upgrade is mandatory IMO, 
LGTM! Thanks David

frelon (approved):
LGTM!

PR #1321, opened 2024-03-26
Fix RT ISO name

davidcassany:


kkaempf (approved):
Hmpf, I _did_ grep for "SLE" in all Dockerfiles before. 🙈 😩 🤦🏻

PR #1323, opened 2024-03-26
Do not start elemental-register timer manually

davidcassany:
Starting the service manually prevent proper ordering and tracking of related services. Having a manual start makes it difficult form other services to check its status as there is no obvious way order the related executions.

Follow up for #1263

### DIFF
--- a/framework/files/usr/lib/systemd/system/elemental-register.timer
+++ b/framework/files/usr/lib/systemd/system/elemental-register.timer
@@ -1,6 +1,9 @@
 [Unit]
 Description=Run elemental-register every 30 minutes
 Documentation=https://elemental.docs.rancher.com
+# Do not run the timer for installation media or recovery system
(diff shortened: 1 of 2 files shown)


davidcassany:
Also requires https://build.opensuse.org/request/show/1161712

framework/files/usr/lib/systemd/system/elemental-register.timer
```diff
@@ -1,6 +1,9 @@
 [Unit]
 Description=Run elemental-register every 30 minutes
 Documentation=https://elemental.docs.rancher.com
+# Do not run the timer for installation media or recovery system
+ConditionPathExists=!/run/elemental/live_mode
+ConditionPathExists=!/run/elemental/recovery_mode
```
davidcassany, line 6: These conditions match with the conditions that were added to the manual start within the deleted file.

frelon (approved):

PR #1378, opened 2024-04-22
ci/ui: increase download time for raw image

juadk:
- Remove boot type raw for RKE2 upgrade because the feature does not exist on Stable
- Create `isBootType` function
- Increase timeout to download the media build

## Verification run
[UI-RKE2-Upgrade](https://github.com/rancher/elemental/actions/runs/8782230157) ✅


frelon (approved):
LGTM!

PR #1382, opened 2024-04-23
ci: enable qase for all tests

### LINKED ISSUE #1379: Enable QASE reporting for daily CI
QASE is used to track where we are with tests and automation. It was asked to us to enable the report for the daily CI to feed internal dashboard:

image

juadk:
Fix #1379 

In the following screenshot, we can see that UI-K3S and Airgap (daily tests) are reported to QASE:

image

I also added variables into the QASE run name to easily identify jobs.


fgiudici (approved):
lgtm

Issue #1362, opened 2024-04-16
[SELinux] RKE2 provisioning


fgiudici:
Check requirements for successful RKE2 provisioning and operations. Check `rke2-selinux` module for targeted policy is enough.

anmazzotti:
I gave this a try building a dev image (TW based) with all latest goodies.

I additionally added the `selinux` flag to the Cluster config. 
This should tell containerd to run in SELinux mode:
```yaml
kind: Cluster
apiVersion: provisioning.cattle.io/v1
metadata:
  name: volcano
  namespace: fleet-default
spec:
  rkeConfig:
    machineGlobalConfig:
      selinux: true
      debug: true
  kubernetesVersion: v1.27.12+rke2r1
```

I installed the `rke2-selinux` package as well.

Also note that rke2 ships `kubectl`, `ctr`, and `crictl` in `/var/lib/rancher/rke2/bin`. See the [docs](https://docs.rke2.io/reference/cli_tools) for usage.

Elemental installation and provisioning of the rke2 node works fine, however rke2 fails to start as the `etcd` container won't start. 
When inspecting it, I found the following:
```text
test-loopdev-c00616c7-e1e9-422c-a786-9bcd826cd8fd:~ # /var/lib/rancher/rke2/bin/crictl logs ecaad625d4c86
{"level":"info","ts":"2024-05-01T09:48:40.917504Z","caller":"etcdmain/etcd.go:73","msg":"Running: ","args":["etcd","--config-file=/var/lib/rancher/rke2/server/db/etcd/config"]}
{"level":"warn","ts":"2024-05-01T09:48:40.917934Z","caller":"etcdmain/etcd.go:75","msg":"failed to verify flags","error":"open /var/lib/rancher/rke2/server/db/etcd/config: permission denied"}
```

```text
test-loopdev-c00616c7-e1e9-422c-a786-9bcd826cd8fd:~ # ls -alZ /var/lib/rancher/rke2/server/db/etcd/config
-rw-------. 1 root root system_u:object_r:var_lib_t:s0 1313 May  1 09:48 /var/lib/rancher/rke2/server/db/etcd/config
```

This seems to be the same issue reported here: https://github.com/rancher/rke2/issues/1494
Which was also reopened as https://github.com/rancher/rke2-selinux/issues/20

anmazzotti:
So maybe I did hit the same error but due to another reason. 
What I noticed is that just after provisioning the labels do not seem to be applied correctly:

```text
[...]

PR #1395, opened 2024-05-02
Set SELinux in permissive mode on recovery, leave default on active

### LINKED ISSUE #1361: [SELinux] K3s provisioning
Check requirements for successfu K3sl provisioning and operations in enforce mode. Check `k3s-selinux` module for targeted policy is enough.

davidcassany:
Fixes #1361
Part of #1362


davidcassany:
I think this is ready to be merged. I manually tested k3s provisioning from an ISO including these changes and I also checked upgrading from staging to this image works and after reboot the system is on enforcing mode.

frelon (approved):
LGTM, but can probably be simplified after https://github.com/rancher/elemental-toolkit/pull/2070 is merged :+1:

davidcassany:
RKE2 provisioning is not fully functional as some files are not getting the appropriate labels. Digging a bit more to figure that out and waiting for https://github.com/rancher/elemental-toolkit/pull/2070 to be merged as this will simplify this PR.

PR #1396, opened 2024-05-03
Drop unused accounting.yaml

frelon:
Very old file which seems to add user root/ros which does not work during live-cd login.


davidcassany (approved):
LGTM

PR #1400, opened 2024-05-06
ci: spread tests over GCP zones

ldevulder:
This should theoretically allow to run more tests in parallel to reduce CI execution time. This PR also fix the scalability and multicluster scheduling.


davidcassany (approved):
I understand the tests are always running on our european night, right?

Daily tests should be triggered at european night as there are less chances to run the during rebuilds triggered by the merge of any of our PRs.

LGTM

ldevulder:
> I understand the tests are always running on our european night, right?

Yes we try. We have to mix with the `us-central1` timezone and our, to have less risk of SPOT and as you said to try to have the tests running during our night.

PR #1404, opened 2024-05-07
Set layout with yaml format

davidcassany:
I prefer having the explicit configuration on yaml and it provides additional configuration possibilities and it is also way more explicit.

This makes it easier to experiment with other setups like having persistent mounted on /var by default, which I think it could be interesting from the SELinux perspective.

Related with #1362

### DIFF
--- a/framework/files/system/oem/01_elemental-rootfs.yaml
+++ b/framework/files/system/oem/01_elemental-rootfs.yaml
@@ -49,7 +49,7 @@ stages:
               - /var/lib/kubelet
               - /var/lib/NetworkManager
               - /var/lib/cni
-              - /var/calico
+              - /var/lib/calico
     - if: '[ -f "/run/elemental/recovery_mode" ]'
       name: "Layout configuration for recovery"
       files: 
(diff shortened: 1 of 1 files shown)


framework/files/system/oem/01_elemental-rootfs.yaml
```diff
@@ -21,32 +21,42 @@ stages:
...
+              - /root
+              - /usr/libexec
+              - /usr/local
+              - /var/log
+              - /var/lib/elemental
+              - /var/lib/rancher
+              - /var/lib/kubelet
+              - /var/lib/NetworkManager
+              - /var/lib/cni
+              - /var/calico
```
fgiudici: is the dropped 'lib' wanted?
```suggestion
              - /var/lib/calico
```

fgiudici (approved):
Wondering if the dropped 'lib' for the calico path is wanted, apart from that looks good!

davidcassany:
> Wondering if the dropped 'lib' for the calico path is wanted, apart from that looks good!

Good catch! thanks much :+1: this was not intended, this was a typo copying paths, thanks!

PR #1405, opened 2024-05-08
Fix grub2-x86_64-efi installation

davidcassany:
Suprisingly enough this has not popped up until late April, when 5.5 images started to fail due to missing grub2-x86_64-efi package.

Apparently this package is usually already installed as part of the previous steps and the explicit installation is usually unneeded in OBS, however for some reason under the stack 5.5 the explicit installation is needed since late April.

### DIFF
--- a/.obs/dockerfile/micro-base-os/Dockerfile
+++ b/.obs/dockerfile/micro-base-os/Dockerfile
@@ -24,7 +24,7 @@ RUN zypper --installroot /osimage in --no-recommends -y filesystem
 RUN zypper --installroot /osimage in --no-recommends -y grub2 shim dracut iputils kernel-default systemd bash
 
 #!ArchExclusiveLine: x86_64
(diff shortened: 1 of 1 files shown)


anmazzotti (approved):
nice one

fgiudici (approved):

kkaempf:
Oh, that one's on me 🤦🏻‍♂️ 
But why didn't this show up earlier ?!

davidcassany:
> Oh, that one's on me 🤦🏻‍♂️ But why didn't this show up earlier ?!

For some reason `zypper install grub2` on the previous step was already resolving to the appropriate `grub2-x86_64-efi` package hence this explicit installation based on the arch was essentially redundant. In fact it is redundant on Micro 6 too, but something changed within grub2/buildenv/zypper in Micro 5.5 that this no longer the case. I am not sure it is worth to track what changed on April 22nd that makes `zypper in grub2` to resolve differently 🤷🏽‍♂️ 

In any case I think this typo needs to be fixed for Micro 6 too, at least to prevent an eventual equivalent issue there.

Issue #1388, opened 2024-04-26
Remove hardcoded OS version for marketplace test


juadk:
Maybe add the os version to install in the `upgrade_from_version` inputs.

https://github.com/rancher/elemental/blob/main/tests/cypress/latest/support/commands.ts#L108-L110

PR #1413, opened 2024-05-14
ci/ui: use OSVersionChannel in marketplace upgrade

### LINKED ISSUE #1389: Improve upgrade test for marketplace
When using k3s in the marketplace upgrade test, `upgrade_image` has to be set to make the k3s upgrade pass.

**For instance:**
`upgrade_image: registry.opensuse.org/isv/rancher/elemental/dev/containers/suse/sl-micro/6.0/baremetal-os-container:latest`

It's hard to guess the full path so maybe we can use OS version channel only when we test the marketplace:

https://github.com/rancher/elemental/blob/main/tests/cypress/latest/e2e/unit_tests/upgrade.spec.ts#L96-L114

juadk:
Fix #1389 and #1388 

# Verification run
[UI-Marketplace-Upgrade-Workflow](https://github.com/rancher/elemental/actions/runs/9113761060) ✅


.github/workflows/sub_test_choice.yaml
```diff
@@ -59,6 +59,12 @@ on:
       os_to_test:
         required: true
         type: string
+      os_version_install:
+        description: OS version to install
```
ldevulder: `description` is only defined in `master-e2e` file (and of course in the `workflow_dispatch` portions too). I did this to avoid the need of adding/modifying/fixing it multiple time, as previously we usually forgot to modify it when needed ;-)

But `required: true` should be added.

.github/workflows/sub_test_choice.yaml
```diff
@@ -59,6 +59,12 @@ on:
...
+        type: string
+      os_version_target:
+        description: OS version to target when upgrading
```
ldevulder: Same for `description` and `required`.

.github/workflows/sub_ui.yaml
```diff
@@ -38,6 +38,12 @@ on:
       operator_repo:
         required: true
         type: string
+      os_version_install:
+        description: OS version to install
```
ldevulder: Same for `description` and `required`.

.github/workflows/sub_ui.yaml
```diff
@@ -38,6 +38,12 @@ on:
...
+        type: string
+      os_version_target:
+        description: OS version to target when upgrading
```
ldevulder: Same for `description` and `required`.

ldevulder (changes requested):
LGTM, just small fixes ;-)

ldevulder (approved):

PR #1434, opened 2024-05-23
Move install openssh and openssh-server

frelon:
Move the installation of openssh and openssh-server to micro-base-os
Dockerfile.

This should include openssh and openssh-server in all derivatives which
is needed for CI.


anmazzotti (approved):
LGTM, makes sense to always have ssh available.

kkaempf:
Uhm. The purpose of `-base` is to provide the _absolute_ minimum. `ssh` is explicitly *not* part of this minimum.

Issue #1450, opened 2024-06-05
`elemental-rt` iso is using the wrong initrd


ldevulder:
Simply try to boot the iso (for example `sl-micro-6.0-rt.x86_64-2.1.1-Build3.33.iso`) and it fails. You can find more information [here](https://openqa.suse.de/tests/14518488#step/installation/4).

kkaempf:
This one seems related to iscsi 🤔

davidcassany:
Found the bug, not related to `iscsi`. The problem is that this image is based on baremetal one, hence it already included an elemental generated initrd for the baremetal image. Initrds are named with `elemental.initrd-<kernel-version>`, the problem is that `zypper in -kernel-default kernel-rt` does not remove any previous initrd and the kernel version is different, hence in `/boot` folder we have two different initrds: `/boot/elemental.initrd-6.4.0-17-default` and `/boot/elemental.initrd-6.4.0-8-rt`. The ISO and grub setup just uses first match of `/boot/elemental.initrd-*`, hence we were using the wrong initrd 🤦🏽 
I am preparing a fix in elemental-toolkit to actually remove any already pre-existing elemental initrd when recreating a new one.

PR #1451, opened 2024-06-06
Workaround to remove pre-existing elemental initrds

davidcassany:
For derived images already including an elemental generated initrd we should remove them specially if the kernel version changes, as the new one would not overwrite the previous one and leave two elemental initrds inside /boot.

Workaround for #1450


kkaempf (approved):
Good catch ! 🤩

PR #1454, opened 2024-06-07
Rebuild kvm initrd

kkaempf:
Same as #1451, but for kvm variant


davidcassany (changes requested):
I guess this PR should only include the last commit. 

@kkaempf I think the PR should be against v2.1.x branch, not to main.

kkaempf:
wrong target branch

PR #1458, opened 2024-06-12
Remove unneeded network setup

davidcassany:
We already removed that from toolkit examples, I don't think this is needed anymore


fgiudici (approved):
LGTM

PR #1464, opened 2024-06-18
elemental.spec: Add procps to Requires

### LINKED ISSUE #1463: oem/03_hostname.yaml requires sysctl
But the respective package (`procps` in (open)SUSE) isn't required and thus missing from the `base` image

kkaempf:
Fixes #1463


davidcassany (approved):
Good catch!

PR #1470, opened 2024-07-01
ci/ui: Remove the x86_64 literal from OS selector

davidcassany:


davidcassany:
@juadk I think this can fix the UI tests as the new OS list in Dev is currently the following 

```
SLE Micro 5.5 ISO v2.0.2-4.2.125
SLE Micro 5.5 ISO v2.0.2-4.2.96
...
SUSE Linux Micro 6.0 ISO (unstable) v2.2.0
```

Machine registration testing -- Create Machine registration we will use to test adding a node (Qase ID 8,20,5) (failed)

This simple hack should solve it for now, but probably instead of having a `contains(<pattern>)` logic we should have `notContains('unstable')` and sort them from high to less, so first (or last) match is always the highest. Do you think this is doable?

davidcassany:
Running a verification [here](https://github.com/rancher/elemental/actions/runs/9745937632) and (rebased) [here](https://github.com/rancher/elemental/actions/runs/9761708941) 🤞🏽

juadk:
> This simple hack should solve it for now, but probably instead of having a `contains(<pattern>)` logic we should have `notContains('unstable')` and sort them from high to less, so first (or last) match is always the highest. Do you think this is doable?

I wanted to have a logic like this to remove the hardcoded value but I have to check how to get all of them and do the sort in my Cypress code.

juadk:
> @juadk I think this can fix the UI tests as the new OS list in Dev is currently the following

I just checked the UI tests and they didn't fail, I still see the architecture in the name, when that change will be applied? I'm confused... 
Today's screenshot with dev:
Screenshot 2024-07-02 at 16 33 13

juadk:
Ok just saw it this morning 😄 
image

juadk (approved):

juadk:
After thinking about the issue, it only happens with rancher 2.7 because we cannot use the marketplace there, that's why we install the latest operator there.

PR #1527, opened 2024-08-06
ci/ui: remove hardcoded version due to new ui rc

juadk:
Rancher `2.9` has its proper version number (2.x) and rancher `2.8` will not see it, so we can remove the harcoded version `1.3.1-rc7` which was only made for `2.9`.

## Verification run
[UI-K3S-rancher2.9-head](https://github.com/rancher/elemental/actions/runs/10268476976) ✅  we can see ui `2.0.0-rc1` installed
[UI-K3S-rancher2.8.5](https://github.com/rancher/elemental/actions/runs/10268507208) ✅  we can see ui `1.3.1-rc9`  installed


kkaempf (approved):
lgtm

PR #1536, opened 2024-08-21
ci/airgap: add custom OS channel

ldevulder:
With operator version 1.7+ there is no default OS channel anymore, so we have to force the use of one (previously only the URL of the registry was changed). The new code is still compatible with older operator versions, as the default one provided in older versions will simply be changed to the new created one.

PR https://github.com/rancher/elemental-operator/pull/824 in the operator repo needs to be merged first!

Verification run:
- [CLI-K3s-Airgap](https://github.com/rancher/elemental/actions/runs/10475126426)


anmazzotti (approved):
LGTM

PR #1537, opened 2024-08-21
ci/cli: reduce execution time on multicluster test

ldevulder:
Due to HW constraints and time consuming the multicluster test sporadically fail. So better to only run it one time instead of 4.

Use K3s for upstream cluster and RKE2 for the downstream ones.


anmazzotti (approved):
LGTM

PR #1539, opened 2024-08-22
ci/openqa: add SLMicro5.5 support

ldevulder:
Support for 5.5 based images needs to be added in openQA update jobgroup to be able to automatically release the new images.


kkaempf (approved):
lgtm

PR #1580, opened 2024-09-20
ci: bump versions of K3s/RKE2

ldevulder:
According to [Rancher support matrix](https://www.suse.com/suse-rancher/support-matrix/all-supported-versions/rancher-v2-9-1/) we should use K3s/RKE2 version 1.27 to 1.30.

Verification run:
- [CLI-K3s-OBS_Dev](https://github.com/rancher/elemental/actions/runs/11028861895)
- [CLI-RKE2-OBS_Dev](https://github.com/rancher/elemental/actions/runs/11031322789)
- [UI-K3s-OBS_Dev](https://github.com/rancher/elemental/actions/runs/11028869269)
- [UI-RKE2-OBS_Dev](https://github.com/rancher/elemental/actions/runs/11033333078)

### DIFF
--- a/.github/workflows/ui-rm-head-2.7-matrix.yaml
+++ b/.github/workflows/ui-rm-head-2.7-matrix.yaml
@@ -14,11 +14,11 @@ on:
         type: boolean
       k8s_downstream_version:
         description: Rancher cluster downstream version to use
-        default: '"v1.27.13+k3s1"'
+        default: '"v1.27.16+k3s1"'
         type: string
       k8s_upstream_version:
         description: Rancher cluster upstream version to use
-        default: '"v1.27.13+k3s1"'
+        default: '"v1.27.16+k3s1"'
         type: string
       proxy:
         description: Deploy a proxy (none/rancher/elemental)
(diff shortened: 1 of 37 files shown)


kkaempf:
Can't we parameterize the k8s version ? Touching 34 files for a (simple) version bump doesn't seem to be very effective.

ldevulder:
> Can't we parameterize the k8s version ? Touching 34 files for a (simple) version bump doesn't seem to be very effective.

It's already  parameterize, matrix and inputs in GitHub Actions cannot be parameterize more unfortunately... We cannot load a file with some parameters for example, which would have been more efficient yes.

But it's not too complicated to change in fact, a simple `sed` command do the magic usually ;-)

.github/workflows/ui-rm-head-2.7-matrix.yaml
```diff
@@ -14,11 +14,11 @@ on:
         type: boolean
       k8s_downstream_version:
         description: Rancher cluster downstream version to use
-        default: '"v1.27.13+k3s1"'
+        default: '"v1.30.4+k3s1"'
         type: string
       k8s_upstream_version:
         description: Rancher cluster upstream version to use
-        default: '"v1.27.13+k3s1"'
+        default: '"v1.30.4+k3s1"'
```
juadk: In rancher 2.7-head, there is nothing higher than 1.27 for the kubernetes downstream cluster:
image

.github/workflows/ui-rm-head-2.7-matrix.yaml
```diff
@@ -42,8 +42,8 @@ jobs:
       max-parallel: 4
       matrix:
         boot_type: ${{ fromJSON(format('[{0}]', inputs.boot_type || '"iso","raw"')) }}
-        k8s_downstream_version: ${{ fromJSON(format('[{0}]', inputs.k8s_downstream_version || '"v1.27.13+k3s1","v1.27.13+rke2r1"')) }}
-        k8s_upstream_version: ${{ fromJSON(format('[{0}]', inputs.k8s_upstream_version || '"v1.27.13+k3s1"')) }}
+        k8s_downstream_version: ${{ fromJSON(format('[{0}]', inputs.k8s_downstream_version || '"v1.30.4+k3s1","v1.30.4+rke2r1"')) }}
+        k8s_upstream_version: ${{ fromJSON(format('[{0}]', inputs.k8s_upstream_version || '"v1.30.4+k3s1"')) }}
```
juadk: Same here for the kubernetes downstream cluster version.

juadk (changes requested):

.github/workflows/ui-rm-head-2.7-matrix.yaml
ldevulder: Yes I saw it, fixed!

.github/workflows/ui-rm-head-2.7-matrix.yaml
ldevulder: Yes I saw it, fixed!

juadk (approved):

PR #1581, opened 2024-09-23
Disable predictable interface names (bsc#1230904)

davidcassany:
This commit adds the net.ifnames=0 kernel parameter to the OS images to prevent interface rename. With the new Micro 6 image, the kernel uses predicable interface naming convention, which causes a network interface rename when upgrading from older OS versions.

This change disables this new kernel feature and keeps the old behavior.

Fixes bsc#1230904


fgiudici (approved):
lgtm

PR #1588, opened 2024-09-24
ci/ui: install operator from cli for ui upgrade

### LINKED ISSUE #1578: Install elemental from CLI instead of using marketplace for upgrade tests
In our UI upgrade tests, we install the stable operator from the marketplace and later, we update it to the latest dev version.
To know which stable ISO version we have to click on, we use a basic [script ](https://github.com/rancher/elemental/blob/main/tests/scripts/start-cypress-tests#L29-L36) but it queries `registry.suse.com`.
And the issue is sometimes, stable from `registry.suse.com` is not the same of the one from the marketplace because these releases are not synced. So the script gets a version which we won't find in the machine registration screen to build an image.

The solution could be to install the stable operator from the commande line instead of using the marketplace and in any case, the installation from the marketplace is tested elsewhere.

## Screenshot of a failed test
image
We can see Cypress is expected to see version `v2.1.2-3.22` in the list but the most updated is `v2.0.4`

juadk:
Fix #1578 

- Install the operator from the command line for upgrade test because registry.suse.com is not always synced with the version we have in the marketplace
- Improve the script to get the latest stable OS version by differentiating Container and ISO

[UI-K3S-Upgrade](https://github.com/rancher/elemental/actions/runs/11028561786/job/30629101422) (test failed for another reason)
We can see that the version matches and the issue does not appear anymore.
image

[UI-K3S](https://github.com/rancher/elemental/actions/runs/11028769489/job/30631744331) ✅ 
[CLI-K3s-OBS_Dev](https://github.com/rancher/elemental/actions/runs/11030568299) ✅


tests/e2e/suite_test.go
```diff
@@ -74,6 +74,7 @@ var (
 	netDefaultFileName        string
 	numberOfClusters          int
 	numberOfVMs               int
+	operatorInstallType       string
 	operatorUpgrade           string
```
ldevulder: You can just fix the `operatorUpgrade` position and put it after `operatorRepo`. Not related to you change and of course no new VR needed ;-)

tests/e2e/suite_test.go
```diff
@@ -757,6 +758,7 @@ var _ = BeforeSuite(func() {
 	k8sUpstreamVersion = os.Getenv("K8S_UPSTREAM_VERSION")
 	number := os.Getenv("VM_NUMBERS")
 	clusterNumber := os.Getenv("CLUSTER_NUMBER")
+	operatorInstallType = os.Getenv("OPERATOR_INSTALL_TYPE")
 	operatorUpgrade = os.Getenv("OPERATOR_UPGRADE")
```
ldevulder: Same comment for `operatorUpgrade` position ;-)

ldevulder (approved):
LGTM, just a small comment not directly related to this PR.

PR #1590, opened 2024-09-24
Adding the new arm-firmware feature for aarch64 builds

### LINKED ISSUE #1516: [PED-8705] Create build-xx-disk-image 'packages'
We already have "build-*-iso-image" packages to create ISO images for the various flavors. See [build-baremetal-iso-image](https://build.opensuse.org/package/show/isv:Rancher:Elemental:Staging/build-baremetal-iso-image) for the baremetal flavor.

PED-8705 now asks for respective 'raw disk' images to simplify OS installation on all supported architectures.

This should be largely similar to the existing build-*-iso-image packages but run "build-disk" instead of "build-iso" when calling `elemental`.

Once done, we can probably drop [elemental-post-build-generate-rpi-image](https://build.opensuse.org/package/show/isv:Rancher:Elemental:Staging/elemental-post-build-generate-rpi-image) as it's no longer needed.

davidcassany:
Requires rancher/elemental-toolkit#2192

Fixes #1516

### DIFF
--- a/.obs/dockerfile/SL-Micro-baremetal-container/Dockerfile
+++ b/.obs/dockerfile/SL-Micro-baremetal-container/Dockerfile
@@ -73,5 +73,9 @@ RUN rm -rf /var/log/update* && \
     >/var/log/lastlog && \
     rm -rf /boot/vmlinux*
 
+# Add the arm-firwmware feature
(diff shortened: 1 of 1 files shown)


frelon (approved):
Pretty clever! :+1:

PR #1596, opened 2024-09-26
ci/ui: fix upgrade cloud-config file

### LINKED ISSUE #1566: Fix update CI test with latest stable operator (1.6.4)
As 1.6.4 is now the new stable, some things can be cleanup up in the Cypress code and some conditions could also be removed.

juadk:
Fix #1566 

As 1.6.4 is the new stable, cloud-config file has to be updated to match SLE Micro 6.

## Verification runs
[UI-K3S-Upgrade](https://github.com/rancher/elemental/actions/runs/11067332822/job/30750705560) ✅

### DIFF
--- a/tests/cypress/latest/fixtures/custom_cloud-config_upgrade.yaml
+++ b/tests/cypress/latest/fixtures/custom_cloud-config_upgrade.yaml
@@ -3,9 +3,34 @@ config:
     users:
     - name: root
       passwd: r0s@pwd1
+    write_files:
+      - path: /etc/ssh/sshd_config.d/root_access.conf
+        append: true
+        content: |
+          PermitRootLogin yes
+        owner: root:root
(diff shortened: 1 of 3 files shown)


tests/cypress/latest/fixtures/custom_cloud-config_upgrade.yaml
```diff
@@ -3,9 +3,34 @@ config:
     users:
     - name: root
       passwd: r0s@pwd1
+    write_files:
+      - path: /etc/ssh/sshd_config
```
ldevulder: If I'm correct the `Stable` version of Elemental OS image are now based on SLMicro6.0 no? In that case the path for sshd config should be `/etc/ssh/sshd_config.d/root_access.conf`.

ldevulder (approved):

PR #1607, opened 2024-10-02
ci/openqa: fix QAM tests

ldevulder:
`KEEP_GRUB_TIMEOUT` variable should be correctly set.

Tested locally on my lab.


kkaempf (approved):
👍🏻

PR #1611, opened 2024-10-03
Merge ISO and OS container images

davidcassany:
This commit unifies OS container and ISO container images at OBS level. This way the two are stick together and we need to handle less submissions.

Also, as a consequence, both release numbers are likely to be aligned with this setup.


.obs/dockerfile/SL-Micro-base-container/Dockerfile.iso
```diff
@@ -10,7 +10,7 @@
 ARG SLMICRO_VERSION
 
 FROM suse/sl-micro/${SLMICRO_VERSION}/base-os-container:latest AS os
-FROM suse/sl-micro/${SLMICRO_VERSION}/baremetal-os-container:latest AS builder
+FROM suse/sl-micro/${SLMICRO_VERSION}/base-os-container:latest AS builder
```
davidcassany, line 13: Using the OS image as the builder itself simplifies the image cross dependencies, so the ISO only requires the OS container to build, so `base` ISO only requires `base` OS.

.obs/dockerfile/SL-Micro-base-container/_service
```diff
@@ -11,4 +11,14 @@
     <param name="eval">IMG_REPO=$(rpm --macros=/root/.rpmmacros -E %img_repo)</param>
     <param name="var">IMG_REPO</param>
   </service>
+  <service name="replace_using_env" mode="buildtime">
+    <param name="file">Dockerfile.iso</param>
+    <param name="eval">SLMICRO_VERSION=$(rpm --macros=/root/.rpmmacros -E %slmicro_version)</param>
+    <param name="var">SLMICRO_VERSION</param>
+  </service>
+  <service name="replace_using_env" mode="buildtime">
```
davidcassany, line 19: nit: we could consider adapting this `replace_using_env` OBS service to actually support globbing, so instead of requiring a new service call per file and per variable we could do something like the following (the `*` on the file name):

```xml
<service name="replace_using_env" mode="buildtime">
    <param name="file">Dockerfile*</param>
    <param name="eval">SLMICRO_VERSION=$(rpm --macros=/root/.rpmmacros -E %slmicro_version)</param>
    <param name="var">SLMICRO_VERSION</param>
  </service>
```

.obs/dockerfile/SL-Micro-baremetal-container/_multibuild
```diff
@@ -0,0 +1,4 @@
+<multibuild>
+  <flavor>iso</flavor>
+  <flavor>disk</flavor>
```
davidcassany, line 3: Note this same approach is also valid for disks

davidcassany (commented):
This PR removes the following OBS packages:


* build-base-iso-image
* build-baremetal-iso-image
* build-baremetal-disk-image
* build-kvm-iso-image
* build-rt-iso-image
* build-k3s-selinux-iso-image
* build-rke2-selinuc-iso-image

And they are embedded in their `SL-Micro-*` counterpart.

davidcassany:
See test build in https://build.opensuse.org/project/show/home:dcassany:Devlinkstest

kkaempf:
I'm not sure about using the OS images `as builder` (which probably contain lots of "build" dependencies, which are not "runtime" dependencies) 🤔

davidcassany:
> I'm not sure about using the OS images `as builder` (which probably contain lots of "build" dependencies, which are not "runtime" dependencies) 🤔

I agree this is arguable, but with the current setup installing `elemental`, which is part of the base image, already includes everything we need to build, which is not that much at the end.

fgiudici (approved):
Looks good here!
Worth just to make the release process easier...  and fixing the mismatch in the release numbers of ISOs and container OS images is quite an important "side effect" too which would alone deserve this PR  💚

PR #1622, opened 2024-10-10
ci/ui: add operator_install_type in rancher 2.7

juadk:
Rancher is installed twice, we have to disable cli installation using `operator_install_type: ui`

## Verification run
https://github.com/rancher/elemental/actions/runs/11269006412


juadk:
Not needed anymore, 2.7 tests are about to be removed.

Issue #1565, opened 2024-09-16
[Epic] Improve OS upgrade logic by using latest features


davidcassany:
By using the latest elemental-toolkit and System Upgrade Controller features we can eventually improve and simplify the suc-upgrade scripts. IMHO we could consider three different improvements:

* Start using the latest elemental-toolkit state.yaml features
* Try to fetch the upgrade OS container from the local storage
* Start using system upgrade controller [exclusive flags](https://github.com/rancher/system-upgrade-controller/pull/260) to prevent concurrent upgrade plans and remove the poor downgrade sanity check

PR #1627, opened 2024-10-11
Use suc-upgrade to invoke elemental-register upgrade

anmazzotti:
Part of https://github.com/rancher/elemental/issues/1565

The entire logic is going to be replaced by `elemental-register upgrade`. See: https://github.com/rancher/elemental-operator/pull/868

The `suc-upgrade` wrapper is still needed for backward compatibility, in case older versions of the `elemental-operator` are tasked to upgrade newer OS images. 

If this is the case, we expect no `ELEMENTAL_REGISTER_UPGRADE_CORRELATION_ID` to be set, and the legacy `UPGRADE_RECOVERY` and `UPGRADE_RECOVERY_ONLY` variables need to be supported.

`FORCE` is no longer supported.

All the other variables were never documented and some of them are hardcoded, so I took the initiative to remove them. 
They can easily be added, as the `elemental-register upgrade` still supports them, like `HOST_DIR` or `CONF_FILE` for example.


ldevulder:
@anmazzotti I think that this PR could be closed, no?

anmazzotti:
Yes, good call. 
This was functional at the time if I recall correctly, so feel free to pick it up again if needed.

PR #1655, opened 2024-11-27
Dockerfile.dev.os: enable elemental-register.timer

fgiudici:
Otherwise elemental-register will not be periodically called (nor will it be started after a reboot).


anmazzotti (approved):
lgtm

PR #1663, opened 2024-12-09
ci/airgap: fix test with Devel version of Rancher Manager

ldevulder:
Verification runs:
- [CLI-K3s-Airgap with Stable and Devel version of Rancher Manager](https://github.com/rancher/elemental/actions/runs/12238352554) :white_check_mark:

**NOTE:** the test on Rncher Manager Alpha version is failing as expected (some dependencies are missing in the `rancher-images.txt` file).


juadk (approved):
LGTM

PR #1681, opened 2025-02-20
ci/ui: use latest stable ISO for maintenance test

juadk:
When we build the ISO, we do not have a case for maintenance.
In maintenance, we want to use the latest stable ISO from the channel brought by the operator maintenance version.

## Verification run
https://github.com/rancher/elemental/actions/runs/13433833637/job/37536844034 ✅


kkaempf (approved):
lgtm

ldevulder (approved):

PR #1683, opened 2025-02-26
Remove /etc/machine-id from base container

frelon:


davidcassany (approved):
LGTM

PR #1686, opened 2025-03-14
Remove /etc/machine-id from base container

davidcassany:
Fixes bsc#1239623

(cherry picked from commit 5dfb218bf35ac3d5af4f62971a74fb969021e6cc)

### DIFF
--- a/.obs/dockerfile/SL-Micro-base-container/Dockerfile
+++ b/.obs/dockerfile/SL-Micro-base-container/Dockerfile
@@ -103,3 +103,5 @@ RUN zypper clean --all && \
 
 # Rebuild initrd to setup dracut with the boot configurations
 RUN elemental init --force elemental-rootfs,elemental-sysroot,grub-config,dracut-config,cloud-config-essentials,elemental-setup,boot-assessment
+
+RUN rm -f /etc/machine-id
--- a/.obs/specfile/elemental/elemental.spec
+++ b/.obs/specfile/elemental/elemental.spec
@@ -1,7 +1,7 @@
 #
 # spec file for package elemental
 #
-# Copyright (c) 2022 - 2024 SUSE LLC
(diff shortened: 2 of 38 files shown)


frelon (approved):

kkaempf:
Why is this required now and wasn't before ? 🤔

frelon:
@kkaempf  A user pointed out that the `/etc/machine-id` was the same for all their machines, which is not expected.

kkaempf:
> @kkaempf A user pointed out that the `/etc/machine-id` was the same for all their machines, which is not expected.

I've seen that. 😉 
I wonder why nobody complained about it before ? Did anything change in SL Micro making `machine-id` appear in the container ?

frelon:
> > @kkaempf A user pointed out that the `/etc/machine-id` was the same for all their machines, which is not expected.
> 
> 
> 
> I've seen that. 😉 
> 
> I wonder why nobody complained about it before ? Did anything change in SL Micro making `machine-id` appear in the container ?

Hehe good question! I did not debug it further than that the file was present (but different) in 5.5, 6.0 and 6.1. The file is not present when building a derivative locally using the toolkit and tumbleweed as base.

itsgottabered:
> > @kkaempf A user pointed out that the `/etc/machine-id` was the same for all their machines, which is not expected.
> 
> I've seen that. 😉 I wonder why nobody complained about it before ? Did anything change in SL Micro making `machine-id` appear in the container ?

I am that user! It's been a few weeks now but I believe 5.5 and 6.0 had an identical machine-id deployed, but nothing really happened because of it. As I recall, the version of systemd in 6.1 changed the default behaviour for MAC generation for bonds/bridges, previously it would adopt the first associated physical adapter but it was changed to fabricate a mac hashed from - you guessed it - /etc/machine-id. Suddenly all my servers had identical macs and dhcp was getting realllllly upset.

PR #1698, opened 2025-04-17
ci/ui: disable container registry for marketplace

juadk:
Remove `Container Registry` code if we test the operator from the marketplace.

## Verification Run
https://github.com/rancher/elemental/actions/runs/14492373441 ✅


ldevulder (approved):
LGTM, thanks!

PR #1727, opened 2025-09-26
ci/cli: use staging in manual upgrade workflow

ldevulder:
Use staging by default, can be changed while triggering the workflow.


davidcassany (approved):
Cool this would also allow to upgrade to an image from a different project. Awesome!

juadk (approved):

PR #1740, opened 2025-12-15
ci: bump helpers librairies

juadk:
Bump libraries to use latest code version.


tests/go.mod
```diff
@@ -1,37 +1,41 @@
 module github.com/rancher/elemental/tests
 
-go 1.24.0
+go 1.24.9
+
+toolchain go1.24.11
```
ldevulder: `toolchain` is not needed.

ldevulder (changes requested):
Just a small comment. And Maybe just change the commit message to `ci: bump librairies` as not only the helpers are bumped.

ldevulder (approved):
LGTM, thanks!

PR #1741, opened 2025-12-30
test: add comprehensive linearizability and race condition tests

### LINKED ISSUE #774: e2e/research: Check if we can use k6 to detect races conditions with elemental-operator API
Check if we can add a test in the CI to monitor elemental-operator API performance and detect regressions etc
It's more a research topic because we are not familiar with the topic, I remember playing a bit with k6 on Epinio but nothing concrete.

copyleftdev:
This PR introduces a comprehensive testing suite for detecting race conditions using Porcupine for linearizability verification and K6 for load testing.

Fixes #774

### Changes
- Added Porcupine dependency.
- Implemented `consistency_test.go` for MachineRegistration annotation linearizability.
- Implemented `inventory_race_test.go` for concurrent MachineInventory creation verified against a Porcupine model.
- Added modular K6 scripts in `tests/e2e/k6/`.
- Validated against both in-memory simulation and real K8s clusters.


ldevulder:
Sorry for the late reply and thanks for you contribution!

But this project does not have lot of new features added as it is in a process to be replaced by a newer version.
On QA side we do not see a huge benefit of having this PR merge, and we also recently had attacks through PRs in other projects so we prefer to not merge it and close this PR.

PR #1742, opened 2026-01-09
set to released-stage to released as SL Micro 6.2 is GA

dirkmueller:


davidcassany:
I think we should apply the change to, at least, rt and kvm flavors too. They are released too.

davidcassany (approved):
LGTM

PR #1750, opened 2026-01-19
ci/cli: fix upgrade test

ldevulder:
We should use `maintenance/6.1` as the "stable" repo and `staging` as the "dev" one.
This is a temporary solution until all maintenance versions are released and a `maintenance/6.2` is created.

Verification runs:
- [CLI-K3-Upgrade](https://github.com/rancher/elemental/actions/runs/21216195287) :white_check_mark:
- [CLI-RKE2-Upgrade](https://github.com/rancher/elemental/actions/runs/21149898506) :x: => It fails because of an issue in Rancher Manager Prime version (should be fixed in the next release).


tests/e2e/upgrade_test.go
```diff
@@ -377,14 +377,7 @@ var _ = Describe("E2E - Upgrading node", Label("upgrade-node"), func() {
 						out = RunSSHWithRetry(cl, "cat /run/elemental/recovery_mode")
 						Expect(out).To(Not(BeEmpty()))
 
-						// Reboot in active (normal) mode
-						_ = RunSSHWithRetry(cl, "grub2-editenv /oem/grubenv set next_entry=active")
```
ldevulder, line 381: Maybe this will have to be re-added, let's see the verification run.

juadk (approved):

tests/e2e/upgrade_test.go
ldevulder: Works as expected, keep it!

PR #1760, opened 2026-02-20
Fix WaitForAllPods and bump backup-restore-operator

ldevulder:
The recently  added `WaitForAllPods` function sporadically fails: this is because the list of pods list is created and then the status checked but meanwhile some pod changed.

This PR fix this by checking pods/ status at once. The PR also outputted pods' status to allow better debugging.

Backup-restore operator is also bumped to v9.0.1 and this version is also used for Rancher Manager v2.14.


davidcassany (approved):
LGTM

PR #1762, opened 2026-02-27
Enable custom channel by default

davidcassany:
Without the channel being enabled the synchronization does not happen and no managedosversions are created. This is relevant for testing specific OS versions in manual test workflows.


ldevulder (approved):
LGTM, thanks!

PR #1773, opened 2026-06-17
ci: Fix gcloud instance create

frelon:
The scheduled tests are failing with the following error:

```
ERROR: (gcloud.compute.instances.create) argument --source-instance-template: expected one argument'
```

The input for runner_template to the action also look empty:

```
 Inputs
    runner_template:
    zone: us-central1-a
```

Looking at [1] it seems the default value is not passed to scheduled
actions. This fix picks the default choice for runner_template for
scheduled actions.

[1] https://stackoverflow.com/questions/73404033/how-to-keep-default-action-input-if-parameter-passed-is-empty-in-github-action


ldevulder (approved):
LGTM, thanks!

